PR #67, opened 2015-01-26
Add the subindexes plugin.

bwinton:
One of the things I missed from PyBlosxom was a sub-index listing all the posts underneath a subdirectory.
So I wrote this.  :)

Any suggestions about how to make it better would be greatly appreciated!

[Review on Reviewable](https://reviewable.io/reviews/getnikola/plugins/67)

### DIFF
--- a/v7/subindexes/subindexes.py
+++ b/v7/subindexes/subindexes.py
@@ -1,5 +1,29 @@
 # -*- coding: utf-8 -*-
 
+# Copyright © 2012-2014 Blake Winton <bwinton@latte.ca>.
+
+# Permission is hereby granted, free of charge, to any
+# person obtaining a copy of this software and associated
+# documentation files (the "Software"), to deal in the
+# Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the
+# Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice
+# shall be included in all copies or substantial portions of
+# the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY
+# KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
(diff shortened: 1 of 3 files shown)


bwinton:
(Uh, that error doesn’t _seem_ like something I caused…)

v7/subindexes/subindexes.py
```diff
@@ -0,0 +1,74 @@
+# -*- coding: utf-8 -*-
```
Kwpolska: missing a license

(can be anything you like)

v7/subindexes/subindexes.py
```diff
@@ -0,0 +1,74 @@
...
+
+from __future__ import unicode_literals
```
Kwpolska: more like

```# Copyright © 2015 Blake Winton.```

Kwpolska:
It looks like this plugin completely ignores archives.  You can easily have a blog which has indexes already there.

For example, my blog has https://chriswarrick.com/blog/2014/ and also https://chriswarrick.com/blog/2014/11/09/inbox-review/ via archives and posts/ structured mirrored in output.

bwinton:
Neat.  How did you get https://chriswarrick.com/blog/2014/11/09/inbox-review/ to show up there?  Is the inbox-review post in a "2014/11/09" directory?

(Also, that seems to solve a slightly different problem than the one I was trying to solve, since I want the sub-indexes to show up, and https://chriswarrick.com/blog/2014/11/ doesn't show all the posts in 2014/11…  :)

Kwpolska:
1. Yes, the path is `posts/2014/11/09/inbox-review.rst` (`/blog/` is handled by `POSTS`)
2. `2014/11` does not work because I didn’t enable `CREATE_MONTHLY_ARCHIVES`, which is overkill for small blogs.  I could also enable `ARCHIVES_ARE_INDEXES` and `CREATE_DAILY_ARCHIVES` for a better effect.
3. Still, your plugin should check that it doesn’t conflict with archives.  Otherwise, a nice error will prevent the site from building due to *two tasks having the same target*.

bwinton:
Ah!  Yes!  I should definitely make it check for that.
Is there a clean way to say "Does this target already exist?" (or "Is another plugin going to handle this target?") so that I can skip generating it?  (That would let me avoid the check at https://github.com/bwinton/plugins/blob/master/v7/subindexes/subindexes.py#L58 too!)

Kwpolska:
No, you’d have to check possibilities manually, according to current archive settings.

PR #68, opened 2015-01-27
Urls equal to Jekyll

magmax:
Generates urls exactly as Jekyll does.

It has still some drawbacks, because I use the "slugged" title and this may be different from what Jekyll does, but it will be closer.

In addition, it imports the option "PRETTY URLS"

I've increased the plugin version too.


Kwpolska:
PS. Nobody cares about plugin version.

PR #69, opened 2015-02-01
Fixing headers.

magmax:
Three small fixes for headers:

1. Teaser mark wasn't being replaced.
2. New lines before the end comment mark of the header may be printed due to a bug in core, but fixed here.
3. Removing quotation in some headers.


Kwpolska:
?

magmax:
Removing one of the changesets that was not necessary.

Sorry, I was in a harry before, so I decided just to close the pull request.

PR #70, opened 2015-02-01
Fixing headers.

magmax:
Two small fixes for headers:

1. Teaser mark wasn't being replaced.
2. New lines before the end comment mark of the header may be printed due to a bug in core, but fixed here.


Kwpolska:
Accept the invite to the organization, this will grant you push rights to this repo.  (it’s rebuilt automatically, we don’t really need to review this)

magmax:
Done. Thank you!

PR #72, opened 2015-02-15
reST HTML5 compiler plugin

pellenilsson:
reST compiler plugin based on this HTML5 writer:

https://bitbucket.org/andre_felipe_dias/rst2html5

Code changes from the original reST plugin are rather trivial, as shown if you diff the two plugins.

Some settings recognised by the HTML4 writer are not recognised by rst2html5: record_dependencies, stylesheet_path and math_output. I haven't investigated the consequences of this (MathJax support might be broken?).

### DIFF
--- a/v7/rest_html5/README.md
+++ b/v7/rest_html5/README.md
@@ -1,0 +1,1 @@
+Modified version of original reST plugin that uses an [alternative docutils writer](https://bitbucket.org/andre_felipe_dias/rst2html5) to produce HTML5 output.
--- a/v7/rest_html5/requirements.txt
+++ b/v7/rest_html5/requirements.txt
@@ -1,0 +1,1 @@
+rst2html5
--- a/v7/rest_html5/rest_html5.plugin
+++ b/v7/rest_html5/rest_html5.plugin
@@ -1,0 +1,10 @@
+[Core]
+Name = rest_html5
+Module = rest_html5
+
+[Documentation]
+Author = Roberto Alsina, Pelle Nilsson
+Version = 1.0
+Website = http://getnikola.com
+Description = Compile reST into HTML5
+
--- a/v7/rest_html5/rest_html5.plugin
+++ b/v7/rest_html5/rest_html5.plugin
@@ -5,6 +5,6 @@ Module = rest_html5
 [Documentation]
 Author = Roberto Alsina, Pelle Nilsson
 Version = 1.0
-Website = http://getnikola.com
+Website = http://plugins.getnikola.com/#rest_html5
 Description = Compile reST into HTML5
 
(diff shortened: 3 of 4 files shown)


ralsina:
record_dependencies is critical, it breaks recompiling the post when included files change and other similar things

pellenilsson:
Actually, the post *is* recompiled if an included file (using the reST include directive) changes when using rest_html5. The dependency seems to be managed by the reST parser in docutils, not the HTML writer (and this seems more natural to me).

Still, the HTML4 writer does update record_dependencies for images and stylesheets, seemingly repeating the code present in the reST parser, but I don't know what this repetition is good for. Do you have a concrete scenario where this causes a problem?

pellenilsson:
I also discovered that rst2html5 doesn't handle HTML comments, and this breaks the TEASER_END mechanism. I'll see if I can get this fixed, so please keep this pull request on hold for now.

andredias:
Could you show me an example of which rst income and html5 outcome would you like from rst2html5? I could implement that.

pellenilsson:
This reST code:

    .. TEASER_END

needs to produce this HTML:

    

With current docutil, it seems that any non-implemented reST directive produces an HTML comment. Nikola relies on this by looking for this HTML comment in a later step of the build process.

pellenilsson:
Oh, and this:

    .. TEASER_END: Any string

also needs to produce

andredias:
ok. I'll do it

andredias:
I've just pushed the implementation at BitBucket. Could you test the development version or do you need me to release a new version? I intend to release it after you test this feature.

pellenilsson:
I've confirmed that your pushed change fixed the TEASER_END issue, thank you very much!

ralsina: Do you still think that missing support for record_dependencies in rst2html5 is a problem?

ralsina:
I am not sure, to be honest :-)

We can always just try it.

andredias:
Please, keep me informed. I am willing to help.

pellenilsson:
As far as I can tell, the plugin is ready. Dependencies on included files seem to work. MathJax works.

I guess we should let @andredias make the new release before merging this, so that the normal `pip install -r requirements.txt` procedure works when this plugin lands on the plugins site.

pellenilsson:
I've also tested most of the old Nikola reST plugins (media, youtube, soundcloud, code-block, chart...)

andredias:
rst2html5 1.5 released!

pellenilsson:
My site that uses this plugin is now live and happy, is there anything I need to do to get this accepted?

Kwpolska:
I guess we could accept this into the repo — we are not forcing this down peoples’ throats yet.  We might, one day.

PR #73, opened 2015-04-03
emojis as a rst transform

mgaitan:
This patch allow a much simpler way to uses emojis. Instead a verbose role, it's use an standard substitution `|emoji_slug|` if that isn't defined in the document and match one of the available emojis


v6/emoji/emoji.py
```diff
@@ -37,13 +38,909 @@ class Plugin(RestExtension):
...
         roles.register_local_role('emoji', emoji_role)
+        if hasattr(site, 'transforms'):
+            site.transforms.append(Emojis)
+        else:
+            site.transforms = [Emojis]
 
 
 def emoji_role(name, rawtext, text, lineno, inliner, options={}, content=[]):
     text = text.lower()
+    LOGGER.warn('The role :emoji:`{0}` is deprecated. Use |{0}| instead'.format(text))
```
Kwpolska: why deprecate it?

mgaitan: because *There should be one-- and preferably only one --obvious way to do it.* (from the Zen)

v6/emoji/emoji.py
```diff
@@ -37,13 +38,909 @@ class Plugin(RestExtension):
...
+
+class Emojis(Transform):
+    """
+    Replace some substitutions if they aren't defined in the document.
+    """
+    # run before the default Substitutions
+    default_priority = 210
+
+    # list from http://www.tortue.me/
+    emojis = set([
```
ralsina: Is there a way to get this list in a file and/or have a way to update it? New emojis happen every once in a while.

mgaitan: Something like 
```
import re
import requests
re.findall(r'"name">(.*)</span', requests.get('http://www.emoji-cheat-sheet.com/').content.decode('utf8'))
```
or watch this project

https://github.com/arvida/emoji-cheat-sheet.com

(but not sure if torue.me is up to date with that one)

v6/emoji/emoji.py
```diff
@@ -37,13 +38,909 @@ class Plugin(RestExtension):
...
+                'bowtie',     # NOQA
```
ralsina: Just curious... why NOQA that one?

mgaitan: strictly, it's over-indented

On Fri, Apr 3, 2015 at 3:14 PM, Roberto Alsina <notifications@github.com>
wrote:

> In v6/emoji/emoji.py
> <https://github.com/getnikola/plugins/pull/73#discussion_r27745195>:
>
> >      node = nodes.image(
> >          uri='http://www.tortue.me/emoji/{0}.png'.format(text),
> >          alt=text,
> >          classes=['emoji'],
> >      )
> >      return [node], []
> > +
> > +
> > +class Emojis(Transform):
> > +    """
> > +    Replace some substitutions if they aren't defined in the document.
> > +    """
> > +    # run before the default Substitutions
> > +    default_priority = 210
> > +
> > +    # list from http://www.tortue.me/
> > +    emojis = set([
> > +                'bowtie',     # NOQA
>
> Just curious... why NOQA that one?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/getnikola/plugins/pull/73/files#r27745195>.
>



-- 
mgaitan.github.io
textosypretextos.com.ar <http://textosyprextextos.com.ar>

Kwpolska:
getnikola/nikola#1647 (required for this) has been merged with some changes: `site.rst_transforms` is used and is guaranteed to exist (no need to do the check).  This plugin will also need to be moved to v7 with MinVersion=7.3.2.

PR #74, opened 2015-04-07
Using th instead of td+b and fixing nested divs

magmax:
Two little improvements:

- I've changed the "td+b" tags by "th" what is more correct, and usually have no visual difference.
- I've fixed the nested "div" tags when no logo is provided.

### DIFF
--- a/v7/projectpages/templates/jinja/project.tmpl
+++ b/v7/projectpages/templates/jinja/project.tmpl
@@ -25,27 +25,27 @@
 
     <table class="table table-hover">
     <tr>
-    <td><b>Status</b></td>
+    <th>Status</th>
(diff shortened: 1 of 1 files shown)


Kwpolska:
Make the same edits in the mako template.

Kwpolska:
Merging.

PS. You pushed your change to getnikola/plugins instead of magmax/plugins.  Also, please add the e-mail associated with that [commit](https://github.com/getnikola/plugins/commit/a3d82e4b18a6088b694affc7e829e2ea9319be41) to github (hover on the ? icon)

magmax:
Sorry, I did it quick and dirty :(
I already fix the commit.
Thank you!

PR #76, opened 2015-04-17
Missing colon

dwilliams:
This caused the plugin to break the building of my site (which at the time had nothing in it).


Kwpolska:
Oh, sorry. Merging, the download will be fixed at midnight UTC.

cc @magmax

PR #79, opened 2015-05-08
New plug-in: Recent posts as JSON

Aeyoun:


v7/recent_posts_json/README.md
```diff
@@ -0,0 +1,22 @@
+This plug-in generates a /index.json file containing the same posts
+as /index.html. Intended to be used in JavaScripts. For example to
+promote the most recent posts at the bottom of older posts.
+
+By default, the JSON file will include:
+    {
+        <JS Date object>: {
+            "title": <post-title>,
+            "iri": <relative-link-to-post>
```
Kwpolska: what does “IRI” mean?

v7/recent_posts_json/README.md
```diff
@@ -0,0 +1,22 @@
+This plug-in generates a /index.json file containing the same posts
+as /index.html. Intended to be used in JavaScripts. For example to
+promote the most recent posts at the bottom of older posts.
+
+By default, the JSON file will include:
+    {
+        <JS Date object>: {
```
Kwpolska: It should be made a bit more open and should be compatible with standard JSON parsers — in other words, use an ISO 8601 date or something more universal

v7/recent_posts_json/README.md
Aeyoun: [Internationalized resource identifier](https://en.wikipedia.org/wiki/Internationalized_resource_identifier).

v7/recent_posts_json/README.md
Aeyoun: Well, … it’s JSON so storing values in JS native formats makes sense, right? Easier to work with from JavaScript. That was my thinking, at least. I usually use ISO 8601 for everything except JSON.

Kwpolska:
@Aeyoun does it work fine for the demo site and your own site?  If yes, we can merge.

Aeyoun:
It works and is deployed on my own sites, and it works on the demo site.

PR #82, opened 2015-05-21
Fix plugins for multi languages

hardening:
The plugin was not migrating the metadata for non default languages, this patch fixes that.

Note: this plugin is not idenpotent if you have a post with default language in new metadata style, and a translation in the old style.

### DIFF
--- a/v7/upgrade_metadata/upgrade_metadata.py
+++ b/v7/upgrade_metadata/upgrade_metadata.py
@@ -69,11 +69,20 @@ class UpgradeMetadata(Command):
                 yesno = utils.ask_yesno("Proceed with metadata upgrade?")
             if options['yes'] or yesno:
                 for post in flagged:
-                    with io.open(post.metadata_path, 'r', encoding='utf-8') as fh:
-                        meta = fh.readlines()
-                    with io.open(post.metadata_path, 'w', encoding='utf-8') as fh:
-                        for k, v in zip(self.fields, meta):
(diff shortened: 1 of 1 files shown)


Kwpolska:
We could just check if the metadata fields match the new syntax.  Adding that and merging.

Kwpolska:
Please do not squash commits!

hardening:
@Kwpolska sorry I was to remove the unused ext var

Kwpolska:
I would have fixed that, too.

v7/upgrade_metadata/upgrade_metadata.py
```diff
@@ -69,11 +69,20 @@ def _execute(self, options, args):
...
-                    with io.open(post.metadata_path, 'r', encoding='utf-8') as fh:
-                        meta = fh.readlines()
-                    with io.open(post.metadata_path, 'w', encoding='utf-8') as fh:
-                        for k, v in zip(self.fields, meta):
-                            fh.write('.. {0}: {1}'.format(k, v))
+                    for lang in post.translated_to:
+                        if lang == post.default_lang or len(post.translated_to) == 1:
+                            fname = post.metadata_path
+                        else:
+                            fname = self.site.config['TRANSLATIONS_PATTERN'].format(path=post.post_name, lang=lang, ext='meta')
```
Kwpolska: It’s better to use `os.path.splitext(post.translated_source_path(lang))` — this is more-or-less what `post.py` does to find the meta files.  Fixed on master.

PR #86, opened 2015-05-30
Deprecate webapp and use it as a Coil advert

Kwpolska:
cc @ralsina

Signed-off-by: Chris Warrick <kwpolska@gmail.com>


ralsina:
+1

PR #87, opened 2015-06-04
Depends on site.timeline

humitos:
Use utils.config_changed({1: self.site.timeline}) as dependecy. So,
when a post is written, changed or deleted the "posts.json" is
re-generated to be up to date.

### DIFF
--- a/v6/random_post/random_post.py
+++ b/v6/random_post/random_post.py
@@ -59,7 +59,7 @@ class Plugin(Task):
                 (create_json, (output_filename, urls)),
                 (create_index, (output_index, )),
             ],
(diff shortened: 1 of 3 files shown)


v6/random_post/random_post.py
```diff
@@ -59,7 +59,7 @@ def gen_tasks(self):
                 (create_json, (output_filename, urls)),
                 (create_index, (output_index, )),
             ],
-            'uptodate': [utils.config_changed(kw)],
+            'uptodate': [utils.config_changed({1: self.site.timeline})],
```
ralsina: Don't remove kw here.

Maybe {1: kw, 2: self.site.timeline} ?

PR #91, opened 2015-06-27
WordPress compiler plugin.

felixfontein:
This plugin allows to compile WordPress posts (i.e. posts directly copied out of a WordPress blog, without any modifications).

It has support for shortcodes. By default, it only supports the [code] shortcode, but plugins for other shortcodes can be added via a plugin interface. See `wordpress/plugins/code.py` for an example.

[Review on Reviewable](https://reviewable.io/reviews/getnikola/plugins/91)

### DIFF
--- a/v7/wordpress_compiler/wordpress/__init__.py
+++ b/v7/wordpress_compiler/wordpress/__init__.py
@@ -21,7 +21,8 @@ from __future__ import unicode_literals
 from .wordpress import CompileWordpress
 from yapsy.IPlugin import IPlugin
 
-__all__ = []
+
+__all__ = ['CompileWordpress', 'WordPressPlugin']
 
 
 class WordPressPlugin(IPlugin):
--- a/v7/wordpress_compiler/wordpress/__init__.py
+++ b/v7/wordpress_compiler/wordpress/__init__.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 # A WordPress compiler plugin for Nikola
 #
 # Copyright (C) 2014-2015 by the contributors
--- a/v7/wordpress_compiler/wordpress/__init__.py
+++ b/v7/wordpress_compiler/wordpress/__init__.py
@@ -2,7 +2,8 @@
 
 # A WordPress compiler plugin for Nikola
 #
-# Copyright (C) 2014-2015 by the contributors
+# Copyright (C) 2014-2015 by Felix Fontein
+# Copyright (C) by the WordPress contributors
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
--- a/v7/wordpress_compiler/wordpress/__init__.py
+++ b/v7/wordpress_compiler/wordpress/__init__.py
@@ -21,4 +21,4 @@ from __future__ import unicode_literals
 from .wordpress import CompileWordpress
 
 
-__all__ = ['CompileWordpress', 'WordPressPlugin']
+__all__ = ('CompileWordpress', )
--- a/v7/wordpress_compiler/wordpress/__init__.py
+++ b/v7/wordpress_compiler/wordpress/__init__.py
@@ -1,0 +1,29 @@
+# A WordPress compiler plugin for Nikola
+#
+# Copyright (C) 2014-2015 by the contributors
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+from __future__ import unicode_literals
+
+from .wordpress import CompileWordpress
+from yapsy.IPlugin import IPlugin
+
+__all__ = []
+
+
+class WordPressPlugin(IPlugin):
+    def register(self, compile_wordpress):
+        pass
--- a/v7/wordpress_compiler/wordpress/default_filters.py
+++ b/v7/wordpress_compiler/wordpress/default_filters.py
@@ -330,13 +330,13 @@ class DefaultWordpressFilters:
         tagregexp = '|'.join([regex.escape(x) for x in self.shortcode_tags.keys()])
 
         pattern = (
-            '<p>'                               # Opening paragraph
+            '<p>'                                 # Opening paragraph
             + '\\s*+'                             # Optional leading whitespace
             + '('                                 # 1: The shortcode
             +     '\\['                           # Opening bracket
             +     "(" + tagregexp + ")"           # 2: Shortcode name
             +     '(?![\\w-])'                    # Not followed by word character or hyphen
-                                                # Unroll the loop: Inside the opening shortcode tag
+                                                  # Unroll the loop: Inside the opening shortcode tag
             +     '[^\\]/]*'                      # Not a closing bracket or forward slash
             +     '(?:'
             +         '/(?!\\])'                  # A forward slash not followed by a closing bracket
--- a/v7/wordpress_compiler/wordpress/default_filters.py
+++ b/v7/wordpress_compiler/wordpress/default_filters.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 # A WordPress compiler plugin for Nikola
 #
 # Copyright (C) 2014-2015 by the contributors
--- a/v7/wordpress_compiler/wordpress/default_filters.py
+++ b/v7/wordpress_compiler/wordpress/default_filters.py
@@ -1,0 +1,362 @@
+# A WordPress compiler plugin for Nikola
+#
+# Copyright (C) 2014-2015 by the contributors
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+# This file is based on WordPress' wp-include/formatting.php
+
+from . import php
+
+import regex
+
+
+def _x(default, lookup):
+    return default
+
+
+class DefaultWordpressFilters:
+    def __init__(self, shortcodes):
+        super().__init__()
+        self.wp_cockneyreplace = None
+        self.__wptexturize_setup()
+        self.__convert_chars_setup()
+        self.option_use_smilies = False  # option!
+        self.wp_smiliessearch = []
+        self.wpsmiliestrans = {}
+        self.shortcode_tags = shortcodes.get_shortcode_tags()
+
+    def __wptexturize_setup(self):
+        # translators: opening curly double quote */
+        opening_quote = _x('&#8220;', 'opening curly double quote')
+        # translators: closing curly double quote */
+        closing_quote = _x('&#8221;', 'closing curly double quote')
+
+        # translators: apostrophe, for example in 'cause or can't */
+        apos = _x('&#8217;', 'apostrophe')
+
+        # translators: prime, for example in 9' (nine feet) */
+        prime = _x('&#8242;', 'prime')
+        # translators: double prime, for example in 9" (nine inches) */
+        double_prime = _x('&#8243;', 'double prime')
+
+        # translators: opening curly single quote */
+        opening_single_quote = _x('&#8216;', 'opening curly single quote')
+        # translators: closing curly single quote */
+        closing_single_quote = _x('&#8217;', 'closing curly single quote')
+
+        # translators: en dash */
+        en_dash = _x('&#8211;', 'en dash')
+        # translators: em dash */
+        em_dash = _x('&#8212;', 'em dash')
+
+        self.default_no_texturize_tags = ['pre', 'code', 'kbd', 'style', 'script', 'tt']
+        self.default_no_texturize_shortcodes = ['code']
+
+        # if a plugin has provided an autocorrect array, use it
+        if self.wp_cockneyreplace is not None:
+            cockney = list(self.wp_cockneyreplace.keys())
+            cockneyreplace = [self.wp_cockneyreplace[key] for key in cockney]
+        elif "'" != apos:  # Only bother if we're doing a replacement.
+            cockney = ["'tain't", "'twere", "'twas", "'tis", "'twill", "'til", "'bout", "'nuff", "'round", "'cause"]
+            cockneyreplace = [apos + "tain" + apos + "t", apos + "twere", apos + "twas", apos + "tis", apos + "twill",
+                              apos + "til", apos + "bout", apos + "nuff", apos + "round", apos + "cause"]
+        else:
+            cockney = []
+            cockneyreplace = []
+
+        static_characters = ['---', ' -- ', '--', ' - ', 'xn&#8211;', '...', '``', '\'\'', ' (tm)'] + cockney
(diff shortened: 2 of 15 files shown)


v7/wordpress_compiler/wordpress/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+from __future__ import unicode_literals
+
+from .wordpress import CompileWordpress
+from yapsy.IPlugin import IPlugin
+
+
+__all__ = ['CompileWordpress', 'WordPressPlugin']
+
+
+class WordPressPlugin(IPlugin):
```
ralsina: Why not make this a PageCompiler? Look at https://github.com/getnikola/nikola/blob/master/nikola/plugins/compile/pandoc.py#L43 for an example.

v7/wordpress_compiler/wordpress/default_filters.py
```diff
@@ -0,0 +1,362 @@
+# A WordPress compiler plugin for Nikola
+#
+# Copyright (C) 2014-2015 by the contributors
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
```
ralsina: It seems to me that this plugin has multiple licenses. It's easier, if it has a GPLv2 part, to just make it all GPL.

ralsina:
Review status: 0 of 10 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

<sup>**[v7/wordpress_compiler/README.md, line 12 \[r1\]](https://reviewable.io:443/reviews/getnikola/plugins/91#-JtE0cYLkepdquiSjD1V)** ([raw file](https://github.com/getnikola/plugins/blob/fde7993e453b7b0e3272861a7071acde9d36c4b8/v7/wordpress_compiler/README.md#L12)):</sup>
If you put this in conf.py.sample it will be automatically shown here, and also shown as a suggestion to the user when he installs the plugin. For example here's what happens when you install markmin:

```
$ nikola plugin -i markmin
INFO:requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): plugins.getnikola.com
[2015-07-02T13:49:30Z] INFO: plugin: Downloading: https://plugins.getnikola.com/v7/markmin.zip
INFO:requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): plugins.getnikola.com
[2015-07-02T13:49:31Z] INFO: plugin: Extracting: markmin into plugins/
[2015-07-02T13:49:31Z] NOTICE: plugin: This plugin has a sample config file.  Integrate it with yours in order to make this plugin work!
Contents of the conf.py.sample file:

    # Add the markmin compiler to your COMPILERS dict.
    COMPILERS["markmin"] = ('.mm',)

    # Add markmin files to your POSTS, PAGES
    POSTS = POSTS + (("posts/*.mm", "posts", "post.tmpl"),)
    PAGES = PAGES + (("stories/*.mm", "posts", "post.tmpl"),)

```

---

<sup>**[v7/wordpress_compiler/wordpress/plugins/code.py, line 3 \[r1\]](https://reviewable.io:443/reviews/getnikola/plugins/91#-JtE1LvuaSeErGxlS1yG)** ([raw file](https://github.com/getnikola/plugins/blob/fde7993e453b7b0e3272861a7071acde9d36c4b8/v7/wordpress_compiler/wordpress/plugins/code.py#L3)):</sup>
Since you wrote it, you shoud use your name in it :-)

---

<sup>**[v7/wordpress_compiler/wordpress/plugins/code.wpplugin, line 2 \[r1\]](https://reviewable.io:443/reviews/getnikola/plugins/91#-JtE1R2tmefdq9_r0ixz)** ([raw file](https://github.com/getnikola/plugins/blob/fde7993e453b7b0e3272861a7071acde9d36c4b8/v7/wordpress_compiler/wordpress/plugins/code.wpplugin#L2)):</sup>
this name is waaaaay too generic and bound to conflict with something.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/getnikola/plugins/91)

v7/wordpress_compiler/wordpress/__init__.py
felixfontein: This is a class which shortcode plugins for the WordPress compiler should derive from. This is *not* the class for the WordPress page compiler itself. The WordPress page compiler is `CompileWordpress` and defined in `wordpress.py`.

ralsina: In any case, inherit at least nikola.plugin_categories.BasePlugin or you will be missing stuff.

v7/wordpress_compiler/wordpress/default_filters.py
felixfontein: I took the original copyright notice found in WordPress source files, and I took their `license.txt` (and added a part for this plugin).

v7/wordpress_compiler/wordpress/__init__.py
felixfontein: Why? The plugin handling for `.wpplugin` plugins (which inherit from `WordPressPlugin`) is done completely by the WordPress page compiler plugin. Nikola won't see these plugins.

v7/wordpress_compiler/wordpress/default_filters.py
ralsina: Oh, it will. Nikola's plugin loader will find them, unless you put them in a different place? Then they will be loaded twice.
Also, there's no reason for the compiler to load plugins, just let Nikola load them and get them from self.site.plugin_manager. Saves you a chunk of code, too.

v7/wordpress_compiler/wordpress/__init__.py
ralsina: Oh, it will. Nikola's plugin loader will find them, unless you put them in a different place? Then they will be loaded twice.
Also, there's no reason for the compiler to load plugins, just let Nikola load them and get them from self.site.plugin_manager. Saves you a chunk of code, too.

v7/wordpress_compiler/wordpress/default_filters.py
ralsina: __init__.py has a MIT-style license, it's easier to just have one.

ralsina: Ah,  my bad, it's also GPL it's just worded differently.

v7/wordpress_compiler/wordpress/__init__.py
felixfontein: No, it won't find them :-) I'm using the plugin extension `.wpplugin`, which Nikola won't look for.

I'll look into using Nikola's `plugin_manager` later.

v7/wordpress_compiler/wordpress/default_filters.py
felixfontein: Are you sure? The only `__init__.py` I can see in this PR has the same GPL boilerplate as all other files.

felixfontein:
> Since you wrote it, you shoud use your name in it :-)

Well, I patched it together, but I took some parts from other people -- like you ;-)

> this name is waaaaay too generic and bound to conflict with something.

Isn't this always local to the `.wpplugin` file?

ralsina:
I try to keep the names consistent, same name in the metadata plugin file, the plugin itself, task names, etc. You never know what someone is going to try in the future :-)

v7/wordpress_compiler/wordpress/__init__.py
felixfontein: I'm not sure whether I can actually use Nikola's plugin manager. When it is constructed (in `Nikola.__init__`, which is before the WordPress plugin is loaded), the plugin categories it should look out for are explicitly specified. So I either have to re-use an existing plugin category, which I think is a really bad idea, or I have to use an own plugin manager (as I'm doing at the moment).

ralsina: Hmmm true. I think that rather points to a problem in our architecture than to a reason to do everything yourself :-)

I could add a "uncategorized" category, and then you could find them via isinstance() or something.

felixfontein: How about a "plugin plugin" category? The base class for that category would have a field for the plugin name this plugin is for, and the Nikola main object would get a function which allows to retrieve all plugins for a given plugin.
I can add something like that to Nikola if you want. The problem is that this won't be part of Nikola anytime soon (7.6.0 was just released), and so the WordPress plugin won't be useable until a newer Nikola version is released -- and in particular, it won't be useable for almost all v7 versions.

felixfontein: Re registering plugin categories from within plugins: that would be really cumbersome, since we have to specify the plugin categories before any plugin is loaded. Except of course if we load plugins *twice*, but I don't really like that idea.

Kwpolska:
```py
class CompilerExtension(BasePlugin):
    compiler_name = ''

class RestExtension(CompilerExtension):
    compiler_name = 'rest'

class MarkdownExtension(CompilerExtension)
    compiler_name = 'markdown'
```

Your plugins would be CompilerExtensions with `compiler_name = 'wordpress'`, and would filter for that.

We could make your plugin v8-only (and we really need to get our act together and land that release) to make compatibility easier.

ralsina:
Ok, let's merge this whenever you want, and let's revisit it in a few months.

v7/wordpress_compiler/wordpress/__init__.py
felixfontein: Yes, that's what I was thinking. Also very good idea to include `RestExtension` and `MarkdownExtension` into this scheme!

Apropos v8: is there a timeline when v8 should be done? For the earlytask branch, the main problem is that it is waiting for some features to appear in official doit before it an be merged.

felixfontein:
I'll add an remark for potential shortcode plugin authors that the plugin interface will be changed for v8, so they should be aware that their plugins have to be changed a bit.
I'll probably manage that by tomorrow.

v7/wordpress_compiler/wordpress/__init__.py
Kwpolska: Well, it will take a while for v8 to become a thing, because there isn’t that much to change yet.  However, I just implemented CompilerExtensions in ea16c26.  You could make your compiler v7.6.1+ and use them.

Kwpolska: Well, it will take a while for v8 to become a thing, because there isn’t that much to change yet.  However, I just implemented CompilerExtensions in getnikola/nikola@c65be6e.  You could make your compiler v7.6.1+ and use them.

felixfontein:
The new version now uses the `CompilerExtension` interface of Nikola 7.6.1.

felixfontein:
I guess the tests will fail until Nikola 7.6.1 is released.

v7/wordpress_compiler/wordpress/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+from __future__ import unicode_literals
+
+from .wordpress import CompileWordpress
+
+
+__all__ = ['CompileWordpress', 'WordPressPlugin']
```
Kwpolska: * should be a tuple, not a list
* where is `WordPressPlugin`?

v7/wordpress_compiler/wordpress/wordpress.py
```diff
@@ -0,0 +1,210 @@
...
+        return self.__uptodate_deps_page
+
+    def __str__(self):
+        return "Context<" + str(self.id) + ">(" + str(self.__file_deps_fragment) + ", " + str(self.__file_deps_page) + ", " + str(self.__uptodate_deps_fragment) + ", " + str(self.__uptodate_deps_page) + ")"
+
+
+class CompileWordpress(PageCompiler):
+    """Compiles a subset of Wordpress into HTML."""
+
+    name = "latex"
```
Kwpolska: Are you sure?

v7/wordpress_compiler/wordpress/wordpress.py
```diff
@@ -0,0 +1,210 @@
...
+from nikola.plugin_categories import PageCompiler
+from nikola.utils import makedirs, write_metadata
+from nikola.utils import get_logger, STDERR_HANDLER
+
+from . import default_filters, shortcodes
+
+_LOGGER = get_logger('compile_wordpress', STDERR_HANDLER)
+
+
+class Context:
```
Kwpolska: `class Context(object):`, use new-style classes

v7/wordpress_compiler/wordpress/wordpress.py
```diff
@@ -0,0 +1,210 @@
...
+        post.add_dependency_uptodate(lambda: self._read_extra_deps(post)[2], True, 'fragment')
+        post.add_dependency_uptodate(lambda: self._read_extra_deps(post)[3], True, 'page')
+
+    def _write_deps(self, context, dest):
+        deps_path = dest + '.dep'
+        if context.has_dependencies():
+            data = (context.get_file_dependencies_fragment(), context.get_file_dependencies_page(),
+                    context.get_uptodate_dependencies_fragment(), context.get_uptodate_dependencies_page())
+            with io.open(deps_path, "wb") as file:
+                pickle.dump(data, file, pickle.HIGHEST_PROTOCOL)
```
Kwpolska: Please use JSON or some sane serialization format. Pickle is arbitrary code execution.

Kwpolska:
Replace ` by the contributors` with `, Felix Fontein.` and add a `# -*- coding: utf-8 -*-` declaration.

To fix the tests, you must duplicate even more code, the plugin test suite uses its own `base.py` file.

v7/wordpress_compiler/wordpress/__init__.py
felixfontein: Thanks! I moved `WordPressPlugin` to `plugin_interface`, otherwise yapsy would (at least for me) load that definition as the plugin and ignore the page compiler...

v7/wordpress_compiler/wordpress/wordpress.py
felixfontein: No :) Thanks for noticing!

v7/wordpress_compiler/wordpress/wordpress.py
felixfontein: Fixed.

felixfontein:
Ok, there's a big problem with the new `CompilerExtensions` interface. How do I now import something from the original WordPress plugin? When it was using its own plugin system, I would just write `import wordpress.xxx` and everything would work fine. Now, I can't anymore.

v7/wordpress_compiler/wordpress/plugins/wordpress_shortcode_code.py
```diff
@@ -0,0 +1,88 @@
...
+import regex
+import threading
+
+import pygments
+import pygments.lexers
+import pygments.formatters
+import pygments.token
+
+
+class UnformattedLexer(pygments.lexer.RegexLexer):
```
Kwpolska: Do you really need that?  There is a `text` lexer available.

Kwpolska:
You could cheat and use `import plugins.wordpress.plugin_interface`, though this will make user-wide plugin installs impossible.

v7/wordpress_compiler/wordpress/wordpress.py
felixfontein: Done.

v7/wordpress_compiler/wordpress/plugins/wordpress_shortcode_code.py
felixfontein: Must have missed that one.

felixfontein:
Doesn't work for me, either, since I'm including the wordpress plugin via `EXTRA_PLUGINS_DIRS`.
I'll Maybe I can expose the modules dynamically in the `register` method. Will be somewhat ugly, but should work. I'll play with that later...

v7/wordpress_compiler/wordpress/plugins/wordpress_shortcode_code.py
felixfontein: Done.

felixfontein:
Hmm, that won't work either, since the first thing I need to do is `import wordpress.plugin_interface` to be able to use `wordpress.plugin_interface.WordPressPlugin`. That sucks bigtime.

Kwpolska:
Hack idea 2: use the plain CompilerExtension class and set `compiler_name = 'wordpress'`.

felixfontein:
And define `register()` myself, and never be able to define any helper functionality in the plugin object itself. Very, very hacky. But probably the only thing that works... :(

Kwpolska:
You could write some functions (≠methods) and pass them to every extension in your compiler, when you register them.

felixfontein:
I'm now providing the modules. I'll do some more testing by porting my other plugins to this way...

felixfontein:
Seems to work this way.

Kwpolska:
Please complete the checklist a few posts above.

felixfontein:
I now also adjusted the copyright lines (though I didn't do precisely the replacement you wanted to do). Is that ok?

felixfontein:
Yeah, just noticed that too :)

felixfontein:
Thanks!

PR #94, opened 2015-08-02
Fotorama directive to display a gallery using http://fotorama.io jQue…

humitos:
…ry plugin

Usage:

.. fotorama::
   :keyboard: true
   :allowfullscreen: native

   image0.jpg
   image1.jpg
   image2.jpg
   image3.jpg


humitos:
For some reason, it is adding `fotorama.js` and `fotorama.css` twice. 

@ralsina @Kwpolska Do you know why? Thanks!

Kwpolska:
I don’t know why that would happen… the hook things haven’t been touched for quite some time. You could just work around it by telling users to change EXTRA_HEAD and BODY_END.

PR #98, opened 2015-08-03
[orgmode] fix bug: wrong translation of characters

lotabout:
'<' '>' '&' characters in src-block are translated to HTML equivalents, which is not expected.


ralsina:
Can't say I can review LISP code :-) Anyone?

PR #111, opened 2015-09-08
Update tipue

ralsina:
localsearch
This PR does the following:

* Update tipue to 5.0 (latest)
* Update produced JSON to work with 5.0
* Nicer example config that shows results in a BS modal popup


ralsina:
Why the heck did I never merge this?

PR #113, opened 2015-09-09
WIP Fix 112

ralsina:
Fix a couple of corner cases found when updating metadata in my blog


ralsina:
Argh, branched off the wrong place.

PR #117, opened 2015-09-18
unicode errors

s2hc-johan:
make decode if we're using python2

### DIFF
--- a/v7/recent_posts_json/recent_posts_json.py
+++ b/v7/recent_posts_json/recent_posts_json.py
@@ -30,6 +30,7 @@ import os
 import io
 import json
 import time
+import sys
 try:
     from urlparse import urljoin
 except ImportError:
@@ -131,7 +132,10 @@ class RecentPostsJon(Task):
             recent_posts.append(entry)
         data = json.dumps(recent_posts, indent=2, sort_keys=True)
         with io.open(output_path, "w+", encoding="utf8") as outf:
-            outf.write(data)
+            if sys.version_info[0] != 3:
+                outf.write(data.decode('utf-8'))
+            else:
+                outf.write(data)
 
     def json_path(self, name, lang):
         if name:  # section
--- a/v7/recent_posts_json/recent_posts_json.py
+++ b/v7/recent_posts_json/recent_posts_json.py
@@ -130,11 +130,11 @@ class RecentPostsJon(Task):
             if previewimage:
                 entry.update({["img"]: post.previewimage()})
             recent_posts.append(entry)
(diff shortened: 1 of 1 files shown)


v7/recent_posts_json/recent_posts_json.py
```diff
@@ -131,7 +132,10 @@ def make_json(self, posts, descriptions, previewimage, output_path, lang):
             recent_posts.append(entry)
         data = json.dumps(recent_posts, indent=2, sort_keys=True)
         with io.open(output_path, "w+", encoding="utf8") as outf:
-            outf.write(data)
+            if sys.version_info[0] != 3:
```
Kwpolska: This is not right.

On Python 2, json.dumps() might return Unicode in some cases. It doesn’t by default; yet we could use to change the default:

```py
data = json.dumps(recent_posts, ensure_ascii=False, indent=2, sort_keys=True)
with io.open(output_path, "w+", encoding="utf-8") as outf:
    try:
        outf.write(data.decode('utf-8'))
    except AttributeError:
        outf.write(data)
````

Comment (s2hc-johan) [inline v7/recent_posts_json/recent_posts_json.py]: Absolutley we can do it like that. Don't know why the first commit is wrong though, isn't json utf-8 by design? In ptyhon2 ".decode('utf-8')" works on both string and unicode objects

Comment (Kwpolska) [inline v7/recent_posts_json/recent_posts_json.py]: It does not work properly.

```pytb
>>> u"ą".decode('utf-8')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u0105' in position 0: ordinal not in range(128)
```

Please switch it to the solution I recommended.

Kwpolska: Actually, in this case we need

```py
except (AttributeError, UnicodeEncodeError, UnicodeDecodeError):

s2hc-johan: yea, excepting more thing makes it more clear.

We don't decode random unicode, we decode output from json.dumps:
    
    >>> import json
    >>> json.dumps([u"ą"]).decode('utf-8')
    u'["\\u0105"]'
    >>>

Kwpolska: Just do it with `ensure_ascii=False`.  More modern.

ralsina:
We no longer care about python 2

PR #118, opened 2015-09-20
Search content must change if any file changed.

punchagan:
The sitemap changes only if a new file has been added, but we can use
these files as dynamic dependencies for the search plugin.


ralsina:
Good idea @punchagan !

PR #119, opened 2015-10-02
Fix to work with Python3

jjconti:
As Nikola is dropping support for Python2 there is no point in having a not merged plugin working on 2 and not on 3.

### DIFF
--- a/v7/nanogallery_directive/templates/mako/embedded-nanogallery.tmpl
+++ b/v7/nanogallery_directive/templates/mako/embedded-nanogallery.tmpl
@@ -3,7 +3,7 @@
 <div id="nanoGallery3">
   % for image in nanogallery_content:
     <a href="${image['href']}" 
-    % for k, v in image['data'].iteritems():
+    % for k, v in image['data'].items():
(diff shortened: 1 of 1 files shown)


v7/nanogallery_directive/templates/mako/embedded-nanogallery.tmpl
```diff
@@ -35,8 +35,8 @@
   // We need window.onload here because jQuery is not loaded yet
   window.onload = function() {
       $("#nanoGallery3").nanoGallery({
-      % for k, v in options.iteritems():
-        % if isinstance(v, unicode):
+      % for k, v in options.items():
+        % if isinstance(v, str):
```
ralsina: If you really want to keep this working in python2, you could use nikola.utils.unicode_str which is always the right thing.

Kwpolska: Is it accessible from within templates?  (and why is there so much logic in this template?)

ralsina:
This PR seems to apply to a plugin that is not there anymore?

PR #121, opened 2015-10-19
Create plugin 'Markdown Extension: Strikethrough'

bameda:
An extension that supports PHP-Markdown style strikethrough.

For example: ``~~strike~~``.


Kwpolska:
Is this really not built-in?  That’s sad.

However, I would rather have that in the Nikola core and not as a plugin.  Could you try adding your code to `nikola/plugins/compile/markdown/mdx_nikola.py` instead?

bameda:
Ok, I can try it.

bameda:
https://github.com/getnikola/nikola/pull/2149

I hope you like it

PR #123, opened 2015-12-04
Added example usage of path handlers from within Asciidoc

otaconix:


Kwpolska:
Merging, thanks!

*(This will go live at midnight UTC)*

PR #126, opened 2016-01-03
template reStructureText directive

humitos:


Kwpolska:
Isn’t this effectively a less functional duplicate of the built-in shortcodes feature that works in reST only?

humitos:
@Kwpolska yes, it seems to cover one of the cases that shortcodes covers. I didn't know about it before @ralsina pointed me to it yesterday.

I don't know how to use shortcodes yet, but it doesn't look to be as easy as including a template via rst directive with meta-template. Isn't it?

BTW, meta-template covers exactly what I needed and only that in a really simple way.

ralsina:
You know what? Let's publish it ayway.

PR #130, opened 2016-01-26
Display a bibtex link for each publication for easy citation.

xuhdev:
This behavior is controlled by the option `:bibtex_dir:`, as explained in README.


Kwpolska:
You can just push to `master` here yourself. The .zip for your plugin will be updated automatically at midnight UTC>

PR #133, opened 2016-02-04
Update orgmode

StreakyCobra:
The first commit corrects the configuration sample, the second update the command used to generate the org-mode syntax highlighting.


Kwpolska:
Thanks!

The changes will be reflected by the Index at midnight UTC.

PR #135, opened 2016-02-20
Added new plugin "issue_role"

LenzGr:
This plugin adds a new reST role `:issue:`, that converts issue IDs to URLs pointing to an issue tracker.

Currently, the issue tracker's URL needs to be configured in the `issue_role.py` plugin file - I have not yet figured out how to obtain this as a configuration option from `conf.py`. Any suggestions on how to achieve this would be appreciated.

Signed-off-by: Lenz Grimmer <lenz@grimmer.com>

### DIFF
--- a/v6/issue_role/README.md
+++ b/v6/issue_role/README.md
@@ -1,0 +1,16 @@
+This plugin adds a new reStructuredText role `:issue:`, which automatically
+creates URLs pointing to the respective issue on the tracker defined in
+`ISSUE_URL`.
+
+Currently, the URL needs to be set in the `ISSUE_URL` variable in
+`issue_role.py`.
+
+`{issue}` will be replaced with the text provided to the `:issue:`
+reStructuredText role.
+
+Example:
+
+When you set `ISSUE_URL` to `http://tracker.yoursite.com/issue/{issue}`
+using ``:issue:`FOO-123` `` in your post will be converted to:
+`<a href="http://tracker.yoursite.com/issue/FOO-123">FOO-123</a>` in the
+resulting HTML page.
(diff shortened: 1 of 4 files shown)


ralsina:
There are a few ways :-)

1) When your plugin's set_site() is called, keep a reference to the site "somewhere"
2) When you need to read the config value, use site.config['NAME_OF_THE_OPTION']

Kwpolska:
Using `site.config` requires [dependency recording](https://getnikola.com/extending.html#restextension-plugins), which might not be easy here. Alternatively, you can use the `self.GLOBAL_CONTEXT` dict, which is you can add to in conf.py (as `GLOBAL_CONTEXT`), which will ensure correct rebuilds.

Kwpolska:
Using `site.config` requires [dependency recording](https://getnikola.com/extending.html#restextension-plugins), which might not be easy here. Alternatively, you can use the `site.GLOBAL_CONTEXT` dict, which is you can add to in conf.py (as `GLOBAL_CONTEXT`), which will ensure correct rebuilds.

LenzGr:
This works for me, but I agree it's somewhat ugly using a global variable here, I admit my Python-fu is not that sophisticated yet. I could not figure out to access the `site` object (and hence the `config` method or `GLOBAL_CONTEXT` dict from within my `IssueRole` function.

Kwpolska:
Use site.GLOBAL_CONTEXT['ISSUE_URL'] and a corresponding entry in your conf.py’s GLOBAL_CONTEXT dict, otherwise you risk rebuild issues.

ralsina:
Merging, I will do the change to GLOBAL_CONTEXT right after that. Thanks for the plugin @LenzGr !!!

LenzGr:
Rocking. Thanks a lot for merging and getting it right - much appreciated.

PR #155, opened 2016-07-01
Use a default directory for projects

punchagan:
Installing the plugin and running any Nikola command before setting the
PROJECT_PATH causes errors.

Also, having a default value for projects is more convenient and in-line
with the defaults we have for posts and pages.


Kwpolska:
That’s not enough. I’ll fix the rest myself and merge.

Kwpolska:
Fixed in commit 1a34ebe.

PR #156, opened 2016-07-22
[blogger] Post type not always be the first tag

punchagan:
Fixes broken import reported by @ChillarAnand [here](https://github.com/getnikola/nikola/issues/408#issuecomment-234269190)


ChillarAnand:
@punchagan Thanks for the quick fix :-)

PR #157, opened 2016-07-24
Preserve blogger directory structure

ChillarAnand:


v7/import_blogger/import_blogger.py
```diff
@@ -191,12 +194,19 @@ def import_item(self, item, out_folder=None):
             # If no content is found, no files are written.
             content = self.transform_content(content)
 
-            self.write_metadata(os.path.join(self.output_folder, out_folder,
-                                             slug + '.meta'),
-                                title, slug, post_date, description, tags)
-            self.write_content(
-                os.path.join(self.output_folder, out_folder, slug + '.html'),
-                content)
+            regex = re.compile('/(?P<year>\d{4})/(?P<month>\d{2})/')
```
Kwpolska: Are there any other URL structures possible with Blogger? This looks like a hack nevertheless. Also, the slug is very likely to still be incorrect (YYYYMMrest-of-slug)

ChillarAnand: This is default structure. Not sure if any other structures are possible.

v7/import_blogger/import_blogger.py
```diff
@@ -191,12 +194,19 @@ def import_item(self, item, out_folder=None):
...
-            self.write_metadata(os.path.join(self.output_folder, out_folder,
-                                             slug + '.meta'),
-                                title, slug, post_date, description, tags)
-            self.write_content(
-                os.path.join(self.output_folder, out_folder, slug + '.html'),
-                content)
+            regex = re.compile('/(?P<year>\d{4})/(?P<month>\d{2})/(?P<slug>.*)')
+            match = regex.search(link_path)
+            if match:
+                year, month, slug = match.group('year'), match.group('month'), match.group('slug')
```
Kwpolska: You don’t need to name your groups in this case and you could just use `match.groups()`

Please check if there are other structures available — and perhaps refrain from using regular expressions, instead basing yourself on splitting on `/`?

v7/import_blogger/import_blogger.py
```diff
@@ -191,12 +194,18 @@ def import_item(self, item, out_folder=None):
...
             content = self.transform_content(content)
 
-            self.write_metadata(os.path.join(self.output_folder, out_folder,
-                                             slug + '.meta'),
-                                title, slug, post_date, description, tags)
-            self.write_content(
-                os.path.join(self.output_folder, out_folder, slug + '.html'),
-                content)
+            link_strings = link_path.split('/')
+            if len(link_strings) == 3:
```
Kwpolska: This is cleaner:

```py
link_fragments = link_path.split('/')
slug = link_fragments[-1]
out_path = os.path.join(self.output_folder, out_folder, link_path)

File: v7/import_blogger/import_blogger.py (hunk 2 above)
Comment (ChillarAnand) [inline v7/import_blogger/import_blogger.py]: Using split seems much simpler. Blogger allows custom slugs but url structure is fixed.

File: v7/import_blogger/import_blogger.py [hunk 4]
```diff
@@ -191,12 +194,16 @@ def import_item(self, item, out_folder=None):
...
             content = self.transform_content(content)
 
-            self.write_metadata(os.path.join(self.output_folder, out_folder,
-                                             slug + '.meta'),
-                                title, slug, post_date, description, tags)
-            self.write_content(
-                os.path.join(self.output_folder, out_folder, slug + '.html'),
-                content)
+            link_fragments = link_path.split('/')
+            if len(link_fragments) == 4:
```
Kwpolska: * this check is unnecessary; however, you should have `slug` pre-determined and set to the correct value much earlier (line 161)
* remove the unnecessary `re` import

v7/import_blogger/import_blogger.py
ChillarAnand: yes please

v7/import_blogger/import_blogger.py
ChillarAnand: What about drafts?

Kwpolska: Detecting drafts does not depend on this as far as I can see. You need to write the correct path to the URL map, for example.

v7/import_blogger/import_blogger.py
```diff
@@ -191,12 +193,17 @@ def import_item(self, item, out_folder=None):
             # If no content is found, no files are written.
             content = self.transform_content(content)
 
-            self.write_metadata(os.path.join(self.output_folder, out_folder,
-                                             slug + '.meta'),
-                                title, slug, post_date, description, tags)
-            self.write_content(
-                os.path.join(self.output_folder, out_folder, slug + '.html'),
-                content)
+            if is_draft:
```
Kwpolska: Why?

v7/import_blogger/import_blogger.py
```diff
@@ -154,8 +156,11 @@ def import_item(self, item, out_folder=None):
 
         if link_path.lower().endswith('.html'):
             link_path = link_path[:-5]
+            link_path = link_path.lstrip('/')
```
Kwpolska: Do this unconditionally.

Kwpolska:
LGTM, but I’d like a second opinion just in case (we went through so many changes here.) @punchagan , @ralsina?

(thanks for listening to all my complaints. In the future, it might be better not to amend and force-push)

v7/import_blogger/import_blogger.py
```diff
@@ -182,21 +187,19 @@ def import_item(self, item, out_folder=None):
         else:
             is_draft = False
 
-        self.url_map[link] = self.context['SITE_URL'] + '/' + \
-            out_folder + '/' + slug + '.html'
+        self.url_map[link] = os.path.join(
```
Kwpolska: You can’t use `os.path.join` for URLs (that would give backslashes on Windows), just use slashes manually or do os.path.join().replace(os.sep, '/')

v7/import_blogger/import_blogger.py
```diff
@@ -182,21 +187,18 @@ def import_item(self, item, out_folder=None):
         else:
             is_draft = False
 
-        self.url_map[link] = self.context['SITE_URL'] + '/' + \
-            out_folder + '/' + slug + '.html'
-
+        self.url_map[link] = self.context['SITE_URL'] + out_folder + \
+                             '/' + link_path + '.html'
```
Kwpolska: better style (flake8 complains about indent):

```py
        self.url_map[link] = (self.context['SITE_URL'] + out_folder +
                              '/' + link_path + '.html')
```

v7/import_blogger/import_blogger.py
```diff
@@ -182,21 +187,18 @@ def import_item(self, item, out_folder=None):
         else:
             is_draft = False
 
-        self.url_map[link] = self.context['SITE_URL'] + '/' + \
-            out_folder + '/' + slug + '.html'
-
+        self.url_map[link] = (self.context['SITE_URL'] + out_folder +
```
punchagan: It might be safer to ensure that SITE_URL ends with a '/' when populating the context.  Some blogs probably don't return a url ending in '/' and that may be why we had it in our code.

v7/import_blogger/import_blogger.py
```diff
@@ -154,8 +156,11 @@ def import_item(self, item, out_folder=None):
 
         if link_path.lower().endswith('.html'):
             link_path = link_path[:-5]
+        link_path = link_path.lstrip('/')
 
-        slug = utils.slugify(link_path)
+        out_path = os.path.join(self.output_folder, out_folder, link_path)
+        link_fragments = link_path.split('/')
+        slug = utils.slugify(link_fragments[-1])
```
punchagan: Are blogger slugs unique?  For example, Could there be posts on different dates `/2016/06/auto-completion-for-custom-search` and `/2014/03/auto-completion-for-custom-search`, which would cause slug clashes.

ChillarAnand: Wont they go to different folders?

v7/import_blogger/import_blogger.py
ChillarAnand: :+1:

v7/import_blogger/import_blogger.py
punchagan: Right.  I was thinking Nikola needed to have unique slugs, but that's not the case. This should be fine.

Kwpolska:
Merging, thanks for your contribution!

Issue #166, opened 2016-09-13
Sass plugin: template inheritance?


felixfontein:
The current implementation of the Sass (and I guess, same for LESS) compiler plugin looks for `sass/targets` in the themes folders, and once it found it, will copy all `.scss` and `.sass` files from the folder where it found the file to `cache/sass`. This implies that you can only use Sass in one theme in the theme inheritance path: only the `.scss`/`.sass` files from the latest theme in the chain are considered.

Wouldn't it make more sense to copy all `.scss`/`.sass` files from all themes to `cache/sass` (resolve collisions by taking the latest) instead?

To implement this, it would make sense to add a functionality to `nikola.utils` which collects all files in a theme subfolder for a chain (with collisions resolved as usual). This could then be used by the Sass and LESS compilers to implement this behavior.

felixfontein:
(I just remembered that the `copy_assets` plugin already does something like that. I'd prefer to change that plugin as well to use a new function in `nikola.utils`, instead of spreading a similar logic into three different plugins.)

ralsina:
I think the less plugin, at least does something like it. It copies all
.less files in the chain into the cache, in fact that's why it uses the
cache.

https://github.com/getnikola/plugins/blob/master/v7/less/less.py#L77

I may be understanding the code wrong, but that's what it seems to be doing
on a quick read.

felixfontein:
Ah, I overlooked that part. That part doesn't handle collisions, though (only the loop above does).

PR #168, opened 2016-09-15
Improve Sass/LESS inheritance

felixfontein:
Fixes problem mentioned in #166

Also improves dependency generation, which currently only considers certain themes (might be related to https://github.com/getnikola/nikola/issues/2495?).


ralsina:
If you say so.... I am not using this code, so, why not. LGTM.

PR #178, opened 2016-10-14
Added hierarchical_pages plugin.

felixfontein:
This is a demonstration how getnikola/nikola#2502 could be used to translate the full URLs of pages without repeating information (like for example the `.. path: xx` solution described in https://github.com/getnikola/nikola/issues/2116#issuecomment-248071723).

PR #179, opened 2016-10-15
Improving hierarchical_pages.

felixfontein:
Part 2 of accidentally merged #178.


Kwpolska:
I’m not sure I understand any of this plugin. This looks like a solution looking for a problem, with 179 lines of code that are just unneeded.

Why are they unnecessary? This is **all** you need to achieve your directory structure, with getnikola/nikola#2502 in place:

```py
PAGES = (
    ("pages/about/team/*.rst", {"en": "about/team", "de": "ueber/mitarbeiter"}, "story.tmpl"),
    ("pages/about/*.rst", {"en": "about", "de": "ueber"}, "story.tmpl"),
    ("pages/*.rst", "", "story.tmpl"),
)
PRETTY_URLS = True
```

Which becomes:

```
(nikola) [kwpolska@Kw-Farore n]% tree pages
pages
├── about
│   ├── company.de.rst
│   ├── company.rst
│   ├── team
│   │   ├── nikola-tesla.de.rst
│   │   ├── nikola-tesla.rst
│   │   ├── roberto-alsina.de.rst
│   │   └── roberto-alsina.rst
│   ├── team.de.rst
│   └── team.rst
├── about.de.rst
└── about.rst

2 directories, 10 files
(nikola) [kwpolska@Kw-Farore n]% tree output/about
output/about
├── company
│   ├── index.html
│   └── index.rst
├── index.html
├── index.rst
└── team
    ├── index.html
    ├── index.rst
    ├── nikola-tesla
    │   ├── index.html
    │   └── index.rst
    └── roberto-alsina
        ├── index.html
        └── index.rst

4 directories, 10 files
(nikola) [kwpolska@Kw-Farore n]% tree output/de/ueber
output/de/ueber
├── firma
│   ├── index.html
│   └── index.rst
├── index.html
├── index.rst
└── mitarbeiter
    ├── index.html
    ├── index.rst
    ├── nikola-tesla
    │   ├── index.html
    │   └── index.rst
    └── roberto-alsina
        ├── index.html
        └── index.rst

4 directories, 10 files

felixfontein:
Yes, you can solve it also like that, but that requires putting all that information into the config! Also, if you use more than one input format, you have to do a lot of repetitions which make it easy to screw something up accidentally.

Having to write everything out is for me the same as having to specify `.. path: xxx` in every post, which is what you didn't like about my original idea (and which I can understand very well).

This plugin allows to specify all paths by using a simple hierarchy in the input folders, and by using the metadata which is already there (i.e. `.. slug: xx`), to automatically construct the translated hierarchy without user intervention or complicated configuration changes.

Kwpolska:
This is not like `.. path:` at all. That meta tag would be repeated all over your posts, while this solution puts them just once in the config. And even a lot of input formats won’t make it that bad, if you can copy-paste stuff right.

This code, as it stands, looks bad. I see this FakePost is needed to use `post.get_meta`, but is that really necessary, including calculating meta twice? Here’s an idea of how to organize things better — and perhaps with less hacks:

1. find only files in pages/*.rst → [about.rst] and turn them into Post objects, save into timeline and hierarchy
2. find only files in pages/about/*.rst → [company.rst, team.rst] → find the directory path by checking the slug for `about` in hierarchy
3. find only files in pages/about/team/*.rst and do the procedure again

Hierarchy in this example could be a dict of `{path: {translations}}` (having the post object here is not even needed). So by the time pages/about/company.rst is found, it would be able to see that 'about' can be translated into 'about' and 'ueber'. The `os.walk` function is perfect for this (other than you’ll have to take care of file extensions yourself)

(I reverted the merge for now, as you might have noticed.)

felixfontein:
Using `Post` to load the metadata is a good idea. I'll implement that. Though without integrating it into the `os.walk` loop.

felixfontein:
Is there any more discussion needed? Can I merge this?

Kwpolska:
I’m not sure if this plugin should exist. @ralsina?

ralsina:
Why not. plugins are by the will of those who do them.

felixfontein:
Thanks for merging!

Issue #176, opened 2016-10-09
[errorpages]


microniko:
I install errorpages plugins but when I compiled the website, I have this error :

> $ nikola build 
> Scanning posts......done!
> Traceback (most recent call last):
>   File "/home/nicolas/envvirt/nikola/lib/python3.5/site-packages/doit/doit_cmd.py", line 168, in run 
>     return command.parse_execute(args)
>   File "/home/nicolas/envvirt/nikola/lib/python3.5/site-packages/doit/cmd_base.py", line 122, in parse_execute
>     return self.execute(params, args)
>   File "/home/nicolas/envvirt/nikola/lib/python3.5/site-packages/doit/cmd_base.py", line 380, in execute
>     self, params, args)
>   File "/home/nicolas/envvirt/nikola/lib/python3.5/site-packages/nikola/**main**.py", line 279, in load_tasks
>     self.nikola.gen_tasks('render_site', "Task", 'Group of tasks to render the site.'))
>   File "/home/nicolas/envvirt/nikola/lib/python3.5/site-packages/doit/loader.py", line 314, in generate_tasks
>     for task_dict, x_doc in flat_generator(gen_result, gen_doc):
>   File "/home/nicolas/envvirt/nikola/lib/python3.5/site-packages/doit/loader.py", line 28, in flat_generator
>     for item in gen:
>   File "/home/nicolas/envvirt/nikola/lib/python3.5/site-packages/nikola/nikola.py", line 1795, in gen_tasks
>     for task in flatten(pluginInfo.plugin_object.gen_tasks()):
>   File "/home/nicolas/envvirt/nikola/lib/python3.5/site-packages/nikola/nikola.py", line 1790, in flatten
>     for ft in flatten(t):
>   File "/home/nicolas/envvirt/nikola/lib/python3.5/site-packages/nikola/nikola.py", line 1789, in flatten
>     for t in task:
>   File "/home/nicolas/repo_site/plugins/errorpages/errorpages.py", line 117, in prepare_error_page
>     task = self.site.generic_renderer(lang, destination, template, self.site.config["FILTERS"], context=context, url_type=url_type)
> AttributeError: 'Nikola' object has no attribute 'generic_renderer'

What happens ?!

felixfontein:
[...]

PR #182, opened 2016-10-23
Adding path handler for errorpages, and setting permalink.

felixfontein:
Fixes plugin part of #176.

(The resource links generated by the default theme are still broken, but that requires Nikola core support.)


Kwpolska:
Looks good, I hope you tested this. Merged.

felixfontein:
Yes, it worked fine here.

PR #185, opened 2016-11-11
orgmode plugin: fix to macros.org (v. 6.0.1)

bennyslbs:
Changed to use https instead of http.
This is needed for the vimeo macro on https pages otherways at least
Firefox blocks http content, on http it should not be an issue to use
https content.

All http:// have been replaced by https://.

Alos replace {vimeo id} with $1, to get correct vimeo url.

### DIFF
--- a/v6/orgmode/orgmode.plugin
+++ b/v6/orgmode/orgmode.plugin
@@ -4,7 +4,7 @@ Module = orgmode
 
 
 [Nikola]
-MinVersion = 6.0.1
+MinVersion = 6.0.0
 plugincategory = Compiler
 
 [Documentation]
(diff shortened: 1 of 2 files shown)


v6/orgmode/orgmode.plugin
```diff
@@ -4,7 +4,7 @@ Module = orgmode
 
 
 [Nikola]
-MinVersion = 6.0.0
+MinVersion = 6.0.1
```
Kwpolska: Don’t change that. This is not plugin version, this is Nikola version.

bennyslbs:
Sorry, I am tired.

Kwpolska:
I’m assuming you didn’t really want to close this PR. Merging.

PR #187, opened 2016-11-16
publication_list: enable multiple highlight authors

cycomanic:
Bibtex files (especially when pulled from different sources) often contain authors in different formats or spelling, for example I have entries with Jochen Schroeder, Jochen Schröder and just initials. This (small) change enables to specify a comma separated list of highlight authors which will be highlighted. I left the option name unchanged (highlight_author, singular) to not break existing sites.

### DIFF
--- a/v7/publication_list/README.md
+++ b/v7/publication_list/README.md
@@ -24,7 +24,7 @@ The `publication-list` directive accepts multiple options.
   default is `papers`.
 
 * `:highlight_author:` indicates the author to highlight. Usually this is the
-  owner of the website. This can be a comma separated list if there are several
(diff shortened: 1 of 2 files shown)


Kwpolska:
Could `,` appear if someone is doing *Lastname, Firstname*? If yes, `;` would work better instead.

Also, please add your `@gmail.com` e-mail address to GitHub settings to ensure proper attribution of commits.

xuhdev:
I agree with @Kwpolska that you probably want a different delimiter.

cycomanic:
Ok, I changed this to semicolon delimited

Kwpolska:
Thanks for contributing to Nikola!

PR #189, opened 2016-11-20
Fixing .dep filename, and adjusting to getnikola/nikola#2547.

felixfontein:
This PR fixes a similar problem as in getnikola/nikola#2550 for the WordPress compiler, adjusts to the changes in PR getnikola/nikola#2547, and also provides a `compile()` function for newest Nikola. This is done (I think) in a backwards-compatible way, so it should work fine with older versions of Nikola.


felixfontein:
Also here, the failing flake8 isn't coming from this PR but from an updated flake8 version.

PR #192, opened 2016-12-03
Notebook shortcode

dwyatte:
This is a small shortcode to adapt existing notebooks into markup-based posts.

### DIFF
--- a/v7/notebook_shortcode/notebook_shortcode.py
+++ b/v7/notebook_shortcode/notebook_shortcode.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+
+# Copyright © 2016 Dean Wyatte
+#
+# Permission is hereby granted, free of charge, to any
+# person obtaining a copy of this software and associated
+# documentation files (the "Software"), to deal in the
+# Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the
+# Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice
+# shall be included in all copies or substantial portions of
+# the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY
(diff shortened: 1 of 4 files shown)


Kwpolska:
Pretty nifty!

Please add this to README:

    `ipynb` must be enabled and configured (COMPILERS, POSTS/PAGES) for CSS to appear properly. If you are using math in your notebook, make sure to add the `mathjax` tag to your post.

Also, This should not leave any `<div id="notebook">` in output (if it does).

v7/notebook_shortcode/notebook_shortcode.py
```diff
@@ -52,8 +52,8 @@ def render_notebook(self, filename, site=None, data=None, lang=None, post=None):
         export_html = HTMLExporter(config=c)
         (notebook_raw, _) = export_html.from_filename(filename)
 
-        # The raw HTML contains garbage (scripts and styles), we can’t leave it in
+        # The raw HTML contains garbage (scripts and styles). Extract only div id=notebook-container and children
```
Kwpolska: `id` won’t be nice either, I’d prefer to avoid having duplicate `id` tags if someone uses multiple notebooks in a post. But it seems you can’t fix it any better.

Kwpolska:
Thanks for contributing! :tada:

The plugin will be available for download after midnight UTC.

v7/notebook_shortcode/notebook_shortcode.py
dwyatte: Seems like we could do something like the following to remove the ids (I'm not too familiar with lxml, so there might be a cleaner way iterate over the notebook elements).

```
notebook_el = notebook_html.xpath('//*[@id="notebook"]')[0]
notebook_el.attrib.pop('id')
notebook_container_el = notebook_html.xpath('//*[@id="notebook-container"]')[0]
notebook_container_el.attrib.pop('id')

notebook_code = lxml.html.tostring(notebook_el, encoding='unicode')
```
I did a quick pass over the ipython css and nikola doesn't do any selection by id, so this appears to work.

At any rate, the merged version also works (at least in my tests), despite not being valid HTML, so we can leave this for the future.

PR #197, opened 2017-01-06
Added static tagcloud plugin.

felixfontein:
Allows to create HTML fragments and CSS for tag clouds which can be included in the blog via JavaScript or some post-processing.

### DIFF
--- a/v7/static_tagcloud/conf.py.sample
+++ b/v7/static_tagcloud/conf.py.sample
@@ -1,0 +1,90 @@
+# Renders two tag clouds for tags and one for categories:
+# - a small one wiht at most 40 tags;
+# - a large one with all tags;
+# - a large one with all categories.
+# 
+# The generated files are
+# - output/tagcloud-LANG.inc
+# - output/tagcloud-LANG-large.inc
+# - output/catcloud-LANG.inc
+# and
+# - output/assets/css/tagcloud-LANG-small.css
+# - output/assets/css/tagcloud-LANG-large.css
+# - output/assets/css/catcloud-LANG-large.css
+
+RENDER_STATIC_TAGCLOUDS = {
+    'tag-small': {
+        # Tag cloud's name (used as CSS class). {0} will be replaced
+        # by the language.
+        'name': 'tcs-{0}',
+
+        # Filename for the HTML fragment. {0} will be replaced by the
+        # language.
+        'filename': 'tagcloud-{0}.inc',
+
+        # The taxonomy type to obtain the classification ("tags")
+        # from.
+        'taxonomy_type': 'tag',
+
+        # Filename for the CSS. {0} will be replaced by the language.
+        'style_filename': 'assets/css/tagcloud-{0}-small.css',
+
+        # Maximum number of levels to be generated
+        'max_number_of_levels': 15,
+
+        # Maximum number of tags in cloud. Negative values mean
+        # that all tags will appear.
+        'max_tags': 40,
+
+        # Tags which appear less often than this number will be
+        # ignored.
+        'minimal_number_of_appearances': 2,
+
+        # Colors defining a gradient out of which the tag font colors
+        # are taken. The colors are specified as RGP triples with each
+        # component being a floating point number between 0.0 and 1.0.
+        'colors': ((0.6,0.6,0.6), (1.0,1.0,1.0)),
+
+        # Colors defining a gradient out of which the tag background
+        # colors are taken.
+        'background_colors': ((0.1, 0.1, 0.1), ),
+
+        # Colors defining a gradient out of which the tag border colors
+        # are taken.
+        'border_colors': ((0.4, 0.4, 0.4), ),
+
+        # Interval (min_value, max_value) for the font size
+        'font_sizes': (6, 20),
+
+        # If positive, will be multiplied by font size to yield the
+        # CSS border radius and the vertical margin. (The horizontal
+        # margin is set to zero.)
+        'round_factor': 0.6,
+    },
+    'tag-large': {
+        'name': 'tcl-{0}',
+        'filename': 'tagcloud-{0}-large.inc',
+        'taxonomy_type': 'tag',
+        'style_filename': 'assets/css/tagcloud-{0}-large.css',
+        'max_number_of_levels': 100,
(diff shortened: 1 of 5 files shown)


v7/static_tagcloud/conf.py.sample
```diff
@@ -0,0 +1,90 @@
+# Renders two tag clouds for tags and one for categories:
+# - a small one wiht at most 40 tags;
```
Kwpolska: with*

v7/static_tagcloud/static_tagcloud/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+
+        # Compose list of tags, their post count and links
+        def acceptor(post):
+            return True if self.site.config['SHOW_UNTRANSLATED_POSTS'] else post.is_translation_available(lang)
+
+        tag_count_url_list = []
+        for tag in natsort.humansorted(list(posts_per_tag.keys())):
+            tag_count_url_list.append((
+                taxonomy.get_classification_friendly_name(tag, lang),
+                len([post for post in posts_per_tag[tag] if acceptor(post)]),
```
Kwpolska: you can easily replace `acceptor` with the inlined logic, which can be simplified to
```py
self.site.config['SHOW_UNTRANSLATED_POSTS'] or post.is_translation_available(lang)
```

v7/static_tagcloud/conf.py.sample
```diff
@@ -0,0 +1,90 @@
...
+
+RENDER_STATIC_TAGCLOUDS = {
+    'tag-small': {
+        # Tag cloud's name (used as CSS class). {0} will be replaced
+        # by the language.
+        'name': 'tcs-{0}',
+
+        # Filename for the HTML fragment. {0} will be replaced by the
+        # language.
+        'filename': 'tagcloud-{0}.inc',
```
Kwpolska: `.inc.html` looks better to me (but feel free to ignore that)

v7/static_tagcloud/static_tagcloud/engine.py
```diff
@@ -0,0 +1,216 @@
...
+        if tag_weight in tag_weight_to_level_weight:
+            level_weight, level_index = tag_weight_to_level_weight[tag_weight]
+            for name, url, index in list:
+                tags.append((index, name, level_weight, level_index, url))
+    tags.sort()
+    return tags, level_weights
+
+
+def _get_hex_color(color):
+    def myHex(i, digits):
```
Kwpolska: String formatting would work better for that, and `str.join` can take a generator expression:

```py
def _get_hex_color(color):
    return '#' + ''.join('{:02x}'.format(int(round(c * 255))) for c in color)
```

v7/static_tagcloud/conf.py.sample
felixfontein: thanks!

v7/static_tagcloud/conf.py.sample
felixfontein: Changed that.

v7/static_tagcloud/static_tagcloud/engine.py
felixfontein: Good idea.

v7/static_tagcloud/static_tagcloud/__init__.py
felixfontein: Done.

Kwpolska:
Merge at will, the site will (or at least, should) update itself on midnight UTC.

felixfontein:
Ok, thanks!

felixfontein:
@Kwpolska: it looks like the plugin site didn't update (correctly).

Kwpolska:
What went wrong?

felixfontein:
The page for sidebar (https://plugins.getnikola.com/#sidebar) still has the broken ReST-style link, and the page for static_tagcloud doesn't seem to exist (https://plugins.getnikola.com/#static_tagcloud).

Kwpolska:
I can see it, force a reload (Ctrl+Shift+R) and blame our single-page architecture.

felixfontein:
Now both vanished for me...

Kwpolska:
Try refreshing [the JS file](https://plugins.getnikola.com/plugin_data.js) or using incognito mode?

felixfontein:
It's really strange. If I load the JS file via `wget` or open it in a new tab, I can see `sidebar` and `static_tagcloud` in there. But when I check out via inspector what the browser is using on the pages, it uses an old version. Forcing reload or using incognito mode in Chromium (I'm testing both Firefox and Chromium) doesn't help, the same thing is happening there. Also, clearing the cache has the same result, which I find even more strange!

felixfontein:
Hmm, maybe this is all related to CloudFlare? They seem to convert JavaScript to "rocketscript", and their rocket scripts are doing the actual loading, so this could be completely independent from what the browser is doing.

felixfontein:
I guess I get the content from another CloudFlare server than you do, and the one I get the content from serves an outdated version of `plugin_data.js`.

Kwpolska:
I tried nuking that file from cache. I should probably get going on the JS-less version of that site.

felixfontein:
https://github.com/Kwpolska/nikola-plugins-v2 is going to be the JS-less version, right?

Kwpolska:
Yes.

felixfontein:
Can I help with that?

Kwpolska:
Sure, I just invited you to that project. I think there isn’t *that* much to do, and to run it, you need to copy `v7` from this repo into the root directory.

felixfontein:
Where does it expects the plugin repository to be?

felixfontein:
Ah. I tried that, but now I get an exception. Looks like one of the plugins is missing a `Nikola` section in the config.

felixfontein:
Actually, that's `v7/notebook_shortcode/notebook_shortcode.plugin`.

felixfontein:
(And is there a better place to discuss this then here?)

Kwpolska:
getnikola/nikola#2482

I fixed `notebook_shortcode.plugin`.

felixfontein:
Ok, thanks!

PR #203, opened 2017-01-10
Simplifying code.

felixfontein:
Simplifies logic and passes more information to page compiler.

Passed language is `site.default_lang` since there are no language-specific comments (yet).


felixfontein:
Thanks for merging.

Kwpolska:
For minor patches to existing plugins, you don’t need a pull request, because the site is automated, and if you made sure it works, I probably can’t improve it.

PR #205, opened 2017-01-14
Added LaTeX formula renderer plugin.

felixfontein:
Allows other plugins to render LaTeX formulae as PNG or SVG images.


v7/latex_formula_renderer/requirements-nopy.txt
```diff
@@ -0,0 +1,11 @@
+a LaTeX distribution::https://en.wikipedia.org/wiki/LaTeX
```
Kwpolska: the file is called nonpy

v7/latex_formula_renderer/latex_formula_renderer.py
```diff
@@ -0,0 +1,787 @@
...
+                return self._convert_to_svg(tempdir, base_name, intermediate_format, intermediate_file, scale, compress=False)
+            elif output_format == 'svgz':
+                return self._convert_to_svg(tempdir, base_name, intermediate_format, intermediate_file, scale, compress=True)
+            else:
+                raise Exception("Unknown output format '{}'!".format(output_format))
+        finally:
+            shutil.rmtree(tempdir, True)
+
+
+class MemoryFile:
```
Kwpolska: Do you really need that? Is io.BytesIO not enough?

felixfontein: Ah, I didn't knew about that one. I'll try that.

PR #206, opened 2017-01-14
Added LaTeX page compiler plugin.

felixfontein:
A page compiler for a subset of LaTeX.

Requires #205.


felixfontein:
(The PR should go to `master` in the end, not necessarily to `latex-formula-renderer`. That is just a "trick" to have both plugins in this branch of the repository to be able to test them together.)

v7/latex/README.md
```diff
@@ -0,0 +1,42 @@
+This plugin allows to write posts and pages in a LaTeX-like syntax.
```
Kwpolska: I really dislike the *-like* part, especially because you did not specify what makes the syntax different from standard LaTeX.

(Pandoc can compile LaTeX to HTML, but the output is pretty ugly there, especially wrt math and spacing)

felixfontein: The idea is that simple enough files (which don't use features the page compiler doesn't support) could be compiled both by the page compiler or being included in a proper LaTeX document and result in the "same" output (as much as HTML and PDF can be the same).

felixfontein: (I tried to make that clearer.)

felixfontein:
I'll add some more documentation later/tomorrow.

Kwpolska:
Merge when you’re done and don’t expect a review.

felixfontein:
Will do.

felixfontein:
Ok, I think I'm done for now. More changes will go into another version.

PR #222, opened 2017-04-02
Fix #221 -- add dependencies in pkgindex_compiler

### LINKED ISSUE #221: [pkgindex_compiler] Updated conf.py.sample files are not updated on plugin site rebuild
Compare https://plugins.getnikola.com/v7/hierarchical_pages/ vs. https://github.com/getnikola/plugins/blob/master/v7/hierarchical_pages/conf.py.sample (I assume you didn't updated the MarkDown by hand...)

Kwpolska:
I’m using `add='both'` implicitly, because some of those files do not appear in `cache/*.html`.

### DIFF
--- a/v7/pkgindex/pkgindex.plugin
+++ b/v7/pkgindex/pkgindex.plugin
@@ -8,6 +8,6 @@ MinVersion = 7.8.2
 
 [Documentation]
 Author = Chris Warrick
-Version = 0.2.1
+Version = 0.2.2
 Website = https://plugins.getnikola.com/
 Description = Generate package indexes (meta-plugin)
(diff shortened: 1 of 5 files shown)


felixfontein:
This is a good start, but there are still things which don't result in proper dependencies, for example `config['demo_screenshots_map']` in `parse_theme_info`: if that changes, no post is being recompiled.

It might be really better to simply do an uptodate dependency in `CompilePackageIndexEntries.read_metadata` on `metadata` returned there, since that implicitly covers all the other added dependencies.

felixfontein:
(And yes, `add='both'` is best here.)

Issue #104, opened 2015-09-01
[orgmode] fontify source block


bricewge:
The present orgmode plugin complain about `htmlize.el` no being present:
```
Cannot fontify src block (htmlize.el >= 1.34 required)
```
Adding something like this to init.el should do the job:
``` emacs-lisp
(require 'package)
(add-to-list 'package-archives '("melpa" . "http://melpa.org/packages/"))
(setq package-load-list '((htmlize t)))
(package-initialize)

(unless (package-installed-p 'htmlize)
  (package-refresh-contents)
  (package-install 'htmlize))
```
Now there is no complain about `htmlize.el` but unfortunatly the HTML produce is broken; here is an example:
``` html
# Code to fontify
(setq bwge/blog-src-directory "~/Dropbox/Blog")
# Without htmlize.el
<div class="highlight"><pre><span class="p">(</span><span class="nf">setq</span> <span class="nv">bwge/blog-src-directory</span> <span class="s">"~/Dropbox/Blog"</span><span class="p">)</span>
# With htmlize.el
<div class="highlight"><pre><span class="p">(</span><span class="nf">setq</span> <span class="nv">bwge/blog-src-directory</span> <span class="nv">&lt;span</span> <span class="nv">style=</span><span class="s">"font-style: italic;"</span><span class="nv">&gt;</span><span class="s">"~/Dropbox/Blog"</span><span class="nv">&lt;/span&gt;</span><span class="p">)</span>

**Comment by bricewge** [comment]
This is caused by the insertion of inline-css code by `htmlize`, which then parsed by `pygments` produce some garbage.
The current workaround that I have found is to set `org-html-htmlize-output-type` to `nil` however this isn't the right way as `htmlize` isn't used anymore and we loose the fontification but not the colorization of the text.

**Comment by punchagan** [comment]
Does the org-file export correctly to html using org-export?
Can you share a sample post with this problem?

**Comment by roblem** [comment]
I am getting this too.  

Emacs 25.1.1
Org 8.2.10
Nikola v7.8.3

Example org file:
```org
#+BEGIN_COMMENT
.. title: Another Test Post using OrgMode Plugin
.. slug: another-test-post-using-orgmode-plugin
.. date: 2017-02-22 21:30:50 UTC+01:00
.. tags: 
.. category: 
.. link: 
.. description: 
.. type: text
#+END_COMMENT

#+BEGIN_SRC python :results output :exports both
print("Hello World")
#+END_SRC
```
yields this in the browser

[...]

PR #223, opened 2017-04-09
Fix #171, #104

### LINKED ISSUE #171: [orgmode] "htmlize.el >= 1.34 required" can be a bogus error
When building a nikola resource with the orgmode plugin I am seeing with what seems to be an error:

```
Cannot fontify src block (htmlize.el >= 1.34 required)
```

This does not prevent the build process to complete -- version of htmlize used is 1.47.

adamcavendish:
The gibberish was generated because `pygmentize` was shown the 'html-formated code', so it cannot correctly highlight the code.

`(code (org-element-property :value src-block))` should be used to directly pull out the source code string out of `src-block`.

And I renamed original `code` to `code-html` to clarify its meaning.

`(code-html (org-html-format-code src-block info))`


Kwpolska:
@punchagan: could you please review this?

adamcavendish:
Add teaser support now.

ralsina:
I actually installed emacs to test this, it LGTM if @punchagan can't get to it soon :-)

PR #224, opened 2017-04-15
Improve latex formula renderer (bugfixes, refactoring, adding support for other TeX engines)

felixfontein:
The new version should be backwards compatible to old instances of the latex compiler plugin (and other plugins).

Also includes a fix for the latex compiler plugin to allow setting the TeX engine for formula rendering.

### DIFF
--- a/v7/latex/latex/plugins/latex_formula_image_renderer.py
+++ b/v7/latex/latex/plugins/latex_formula_image_renderer.py
@@ -89,6 +89,7 @@ class LatexImageFormulaRenderer(nikola.plugin_categories.CompilerExtension):
         super(LatexImageFormulaRenderer, self).__init__()
         self.__formula_scale = 1.25
         self.__formula_color = (0., 0., 0.)
+        self.__formula_engine = 'latex'
 
     def _get_formulae_filename(self, post, lang):
         """Get filename for post and language to store LaTeX formulae in."""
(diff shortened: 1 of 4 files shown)


v7/latex/latex/plugins/latex_formula_image_renderer.py
```diff
@@ -89,6 +89,7 @@ def __init__(self):
         super(LatexImageFormulaRenderer, self).__init__()
         self.__formula_scale = 1.25
```
Kwpolska: PS.  I really dislike those double underscores.

v7/latex_formula_renderer/latex_formula_renderer.py
```diff
@@ -43,7 +43,171 @@
...
+        'steps': [_create_latex_call('pdflatex', 'pdf')],
+    }],
+    ######################################################################
+    'luatex': [{
+        'types': {'display', 'inline', 'tikzpicture'},
+        'texfile': {
+            'programs': ['lualatex'],
+            'head': R"""\RequirePackage{luatex85}
+\documentclass{standalone}
+\usepackage[T1]{fontenc}
```
Kwpolska: `\usepackage{fontspec}` would be a better choice for LuaTeX.

Kwpolska:
(Merge at will, since plugins do not *require* reviews.)

v7/latex_formula_renderer/latex_formula_renderer.py
felixfontein: Good idea.

v7/latex/latex/plugins/latex_formula_image_renderer.py
felixfontein: Because they look ugly, because they don't really give privacy, or because of something else?

Kwpolska: That, and they may mess up in some situations where you never expect it. This probably won’t happen in a Nikola plugin that doesn’t get inherited though.

PR #229, opened 2017-05-21
Initial JSON Feed implementation

Kwpolska:
Fixes getnikola/nikola#2779.

An experimental implementation of the [JSON Feed](https://jsonfeed.org/) specification (version 1).

Currently only supports main blog RSS (``/feed.json``). Other items are not supported, but are planned for the future.

- [x] tags → ~~missing for untranslated tags~~
- [x] categories
- [x] sections
- [ ] ~~galleries~~ (painful without changes in Core?)
- [x] archives (WIP; getnikola/nikola#2783)
- [x] authors
- [x] add links (JSONFEED_APPEND_LINKS)

Comments, ideas, insights welcome. (I had to make some uglier architectural decisions, including duplicating code, because this is not in Core.)

### DIFF
--- a/v7/jsonfeed/README.md
+++ b/v7/jsonfeed/README.md
@@ -1,0 +1,5 @@
+An experimental implementation of the [JSON Feed](https://jsonfeed.org/) specification (version 1).
+
+Currently only supports main blog RSS (``/feed.json``). Other items (tags,
+categories, sections, galleries, archives, authors, ?) are not supported but
+are planned for the future.
--- a/v7/jsonfeed/README.md
+++ b/v7/jsonfeed/README.md
@@ -1,5 +1,5 @@
 An experimental implementation of the [JSON Feed](https://jsonfeed.org/) specification (version 1).
 
 Currently only supports main blog RSS (``/feed.json``). Other items (tags,
-categories, sections, galleries, archives, authors, ?) are not supported but
+categories, sections, galleries, archives, authors, ?) are not supported, but
 are planned for the future.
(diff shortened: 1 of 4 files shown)


felixfontein:
BTW, how about a bit of infrastructure in Nikola core which allows to add new feed types (or remove existing ones by disabling corresponding plugins)? Currently, RSS and Atom are kind of hard-coded and it is hard to change their behavior.

Also, Atom and RSS behave very differently: Atom is like another representation of the accompanying HTML page, while RSS is what I traditionally knew as a feed. I guess the infrastructure would have to support both modes. Or are there even more modes than these two to consider?

Kwpolska:
How many feed types are there, really? I don’t think we need to bother. This plugin works pretty well without any core changes, the only things I can’t handle very nicely from here would be galleries (but it’s doable with some copy-paste, I’ll see if it’s doable later).

@felixfontein There is an issue with tags: site.posts_per_classification does not have untranslated tags. Those tags *do* get RSS feeds and indexes, should I use something else?

felixfontein:
This also happens for categories, and it is some kind of left-over from the past / something for compatibility to the old version.

There's some extra code in `taxonomies.py` which collects tag and category names from all languages when `also_create_classifications_from_other_languages` is `True`, which happens precisely for the tag and category taxonomies. I guess you have to copy that behavior.

(I'd personally would be happy if we'd remove this behavior from the taxonomies at some point in the future. I don't see any reason why these should be generated in the first place.)

felixfontein:
About feed types: no idea. I'm happy with RSS and Atom (though I still don't know what Atom is for anyway). It would just also clean up the RSS and Atom code a bit, currently explicit RSS and Atom handling is all over the place.

Kwpolska:
This is pretty done now, there seem to be some dependency problems (probably caused by template_handlers). I’ll look at it tomorrow.

@felixfontein make an issue in the main repo for `also_create…` and other things in Taxonomies that should be deprecated/changed in v8.

PR #230, opened 2017-05-22
Similarity plugin (WIP)

ralsina:
https://github.com/getnikola/nikola/issues/1702

### DIFF
--- a/v7/similarity/README.md
+++ b/v7/similarity/README.md
@@ -11,8 +11,7 @@ For example, if your post is in `output/foo/bar.html` then
 the related posts data will be in `output/foo/bar.html.related.json`
 and looks like this:
 
-```
-[
+```[
   {
     "detailed_score": [
       0,
@@ -43,4 +42,33 @@ Caveats:
 * You will need to hack your templates to load the
   similarity data and display it to the user.
 
-**TODO** Add example JS to load and display data
+Assuming you use a theme that makes JQuery available like bootstrap3,
+this would work. If not, change as needed.
+
+In post.tmpl, where you want the related post links to appear:
+
+```html
+    <div id="related-posts" class="related">
+    <h3>Related Posts:</h3>
+    </div>
+```
+
+And then add a script to load them there, like this:
+
+```
+<%block name="extra_js">
+    <script>
+        jQuery.getJSON("${post.permalink()}.related.json", null, function(data){
+            var items = [];
+            $.each(data, function(i) {
+                items.push("<li><a href="+data[i].url+">"+data[i].title+"</a></li>")
+            });
+            $( "<ul/>", {
+                "class": "related-items",
+                html: items.join( "" )
+            }).appendTo( "#related-posts" );
+        });
+    </script>
+</%block>
+
+```
--- a/v7/similarity/README.md
+++ b/v7/similarity/README.md
@@ -1,0 +1,46 @@
+A plugin to create "related posts" data.
+
+This plugin uses [gensim](https://radimrehurek.com/gensim/) to
+analyze your posts and find which ones are similar to one another.
+
+The scoring mechanism is "inspired" by [YARPP](https://www.slideshare.net/mitcho/the-yet-another-related-posts-plugin-algorithm-explained/9)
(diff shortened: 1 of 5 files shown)


v7/similarity/similarity.py
```diff
@@ -0,0 +1,61 @@
...
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+# PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS
+# OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+# OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+# SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+from __future__ import print_function, unicode_literals
+
+import gensim
```
Kwpolska: needs a requirements.txt file

v7/similarity/similarity.py
```diff
@@ -0,0 +1,61 @@
...
+        lsi = gensim.models.LsiModel(corpus, id2word=dictionary, num_topics=2)
+        index = gensim.similarities.MatrixSimilarity(lsi[corpus])
+
+        for i, post in enumerate(self.site.timeline):
+            doc = texts[i]
+            vec_bow = dictionary.doc2bow(doc)
+            vec_lsi = lsi[vec_bow]
+            sims = index[vec_lsi]
+            sims = sorted(enumerate(sims), key=lambda item: -item[1])
+            print(i, sims[:10])
```
Kwpolska: Hm, shouldn’t this write to eg. a JSON file? Or otherwise make available to others?

Also, you have no doit tasks here.

ralsina: That was very early code :)

ralsina: That was very early code :)

v7/similarity/similarity.py
```diff
@@ -0,0 +1,124 @@
...
+                title_sims = [title_similarity(post, p) for p in self.site.timeline]
+                full_sims = [tag_sims[i] + title_sims[i] + body_sims[i] * 1.5 for i in range(len(self.site.timeline))]
+                full_sims = sorted(enumerate(full_sims), key=lambda item: -item[1])
+                related = [(self.site.timeline[s[0]], s[1], tag_sims[s[0]], title_sims[s[0]], body_sims[s[0]]) for s in full_sims[:11] if s[0] != i]
+                task = {
+                    'basename': self.name,
+                    'name': out_name,
+                    'targets': [out_name],
+                    'actions': [(write_similar, (out_name, related))],
+                    # 'file_dep': ['####MAGIC####TIMELINE'],
```
Kwpolska: I think you need to depend on the `render_posts` task for magic timeline to work.

ralsina: ahhhh good point. I sort of need to do it differently, so I may have to extend that support.

v7/similarity/README.md
```diff
@@ -0,0 +1,46 @@
...
+Caveats:
+
+* This will make your builds slower.
+
+  * There is a considerable and somewhat unavoidable startup cost.
+  * Any change in any post involves recalculating the
+    similarity data for **all** posts.
+  * The more translations you have, the longer it takes.
+  * My test site contains 1300 posts of varied lengths in 2 languages,
+    and initialization takes ~90 seconds.
```
Kwpolska: * Is it really worth it?
* How powerful is your machine?
* Is there a way to save the startup data? Even with something like manual pickle/unpickle?

ralsina: Yes, there is, and I can make this better using setup tasks, which is a feature in doit I did not know/recall.

Kwpolska:
The $64,000 question is: how can WordPress plugins generate a list like that dynamically on every request? Perhaps this approach is a bit overdone?

ralsina:
Oh, they have all the text in a DB with FTS. With that, it's easier.

PR #231, opened 2017-05-27
Continuous import

ralsina:
Merging even as WIP


v7/continuous-import/README.md
```diff
@@ -0,0 +1,31 @@
+The continuous import plugin lets Nikola work as a hub for the different
```
Kwpolska: Using an underscore in the plugin directory name would be better.

ralsina: ok, changing

PR #243, opened 2017-07-26
Localsearch non root context

### LINKED ISSUE #242: [localsearch]: Works only for nikola within the root context
* OK: SITE_URL="http://my-server.com/"
* KO: SITE_URL="http://my-server.com/nikola/"

uli-heller:
Fixes #242.

Please note: I reordered conf.py.sample to match the ordering within conf.py. This is not required for #242.

Related to #242: It is important to prepend the SITE_URL when loading via javascript. My python skills
are very low. I didn't succeed in adding a `.format()` to the complete BODY_END, maybe since there are
extra brackets within the string. So I splitted BODY_END into parts and used `format()` only on the middle
part.

### DIFF
--- a/v7/localsearch/conf.py.sample
+++ b/v7/localsearch/conf.py.sample
@@ -32,17 +32,22 @@ BODY_END = """
   </div>
 </div>
 <script>
+"""
+BODY_END += """
+    var siteUrl = "{siteUrl}"
+""".format(siteUrl = SITE_URL)
+BODY_END += """
 $(document).ready(function() {
     $.when(
-        $.getScript( "/assets/js/tipuesearch_set.js" ),
-        $.getScript( "/assets/js/tipuesearch.js" ),
+        $.getScript( siteUrl + "/assets/js/tipuesearch_set.js" ),
+        $.getScript( siteUrl + "/assets/js/tipuesearch.js" ),
         $.Deferred(function( deferred ){
             $( deferred.resolve );
         })
     ).done(function() {
         $('#tipue_search_input').tipuesearch({
             'mode': 'json',
-            'contentLocation': '/assets/js/tipuesearch_content.json'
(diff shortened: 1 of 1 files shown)


v7/localsearch/conf.py.sample
```diff
@@ -28,17 +32,22 @@ BODY_END = """
   </div>
 </div>
 <script>
+"""
+BODY_END += """
```
Kwpolska: Do it in one assignment, using string concatenation instead. Also, use `json.dumps()` to escape the URL properly (adds quotes).

uli-heller:
Am 26.07.2017 um 16:51 schrieb Chris Warrick:
> Do it in one assignment, using string concatenation instead.

I've done this.

> Also, use |json.dumps()| to escape the URL properly (adds quotes). 
I don't get this. Do you mean

* I should use `...+json.dumps(SITE_URL)+...`
* instead of `...+SITE_URL+...` ?

Does this work within conf.py? I'm not sure where to put the `import 
json` within
this file.

Thanks + best regards, Uli

-- 
Uli Heller

daemons point Gesellschaft für Datenverarbeitung mbH
Zahn-Nopper-Straße 1-3, 70435 Stuttgart
Internet: www.daemons-point.com
Mobil: +49 17273 05646, Tel: +49 711 806708 0, Fax: +49 711 806708 19
mailto:uli.heller@daemons-point.com

Kwpolska:
Put `import json` at the top of your conf.py.sample file, and then use:

    var siteUrl = """ + json.dumps(SITE_URL) + """

(note the removed single quotes)

uli-heller:
Done. Thanks a lot for your advice.

Kwpolska:
Thanks for fixing this!

uli-heller:
Thanks for pulling!

PR #248, opened 2017-09-15
Adapted to Nikola v8. Changed some incompatible legends

marian-vignau:
Fix bugs that prevent plugin from render, changing "Recent Posts" to "Newer posts" and "Archives" to "Archive"


felixfontein:
For me, the v7 plugin was working fine with Nikola v8.

Kwpolska:
I ran a `diff`, only the templates were changed to substitute messages. As such, we have to refuse this change — the Nikola v7 plugin works fine, as long as you read the README file and add the missing translations by yourself.

felixfontein:
Or use a different (customized) template. The included ones are only examples.

PR #251, opened 2018-01-07
Making hierarchical_pages v8 ready (new metadata scanning).

felixfontein:
The only difference to the v7 version is `metadata_extractors_by=self.site.metadata_extractors_by` in the `Post` constructor. (And adjusting the documentation.)


felixfontein:
Thanks for merging!

PR #257, opened 2018-03-08
[WIP] add new uuid-based shortcodes for the v8 compiler plugins that lack it

ralsina:
* new shortcode support for v8/kramdown
* wiki (mediawiki) plugin is broken because smc.mw is broken, so deleted
* txt2tags is py2-only and v8 is py3-only so deleted


ralsina:
I think this is all that can be done, details at https://github.com/getnikola/nikola/issues/2752

PR #268, opened 2018-04-19
Add plugin to create crumbs based on metadata

tbm:
The `pretty_crumbs` plugin takes metadata information from the source
file (by default, the `crumbs` variable but this can be configured) and
uses that information for the breadcrum instead of the file/dirname.

Thanks to Chris Warrick for answering several questions I had as I
worked on this plugin.

### DIFF
--- a/v8/pretty_breadcrumbs/README.md
+++ b/v8/pretty_breadcrumbs/README.md
@@ -1,0 +1,16 @@
+
+This plugin converts filepaths from breadcrumbs into more pretty
+breadcrumbs based on metadata information defined in the source
+file.
+
+For example, if file `test/index.rst` contains
+
+```
+.. crumb: This is a test
+```
+
+the breadcrumb generated would be `This is a test` instead of `test`.
+
+By default, the plugin looks for the metadata tag `crumb` but this
+can be changed via `PRETTY_BREADCRUMBS_TAG`.
+
--- a/v8/pretty_breadcrumbs/conf.py.sample
+++ b/v8/pretty_breadcrumbs/conf.py.sample
@@ -1,0 +1,5 @@
+
+# By default, the pretty_crumbs plugin looks for the metadata tag `crumb`.
+# This can be changed with PRETTY_BREADCRUMBS_TAG.
+#PRETTY_BREADCRUMBS_TAG = "nav"
+
--- a/v8/pretty_breadcrumbs/pretty_breadcrumbs.py
+++ b/v8/pretty_breadcrumbs/pretty_breadcrumbs.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+
+# Copyright © 2018 Martin Michlmayr
+
+# Permission is hereby granted, free of charge, to any
+# person obtaining a copy of this software and associated
+# documentation files (the "Software"), to deal in the
+# Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the
+# Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice
+# shall be included in all copies or substantial portions of
+# the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY
+# KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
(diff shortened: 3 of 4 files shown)


tbm:
This is my first plugin so feedback appreciated.

It seems crumbs.tmpl was renamed to breadcrumbs.tmpl so I think I should change the name of the plugin to pretty_breadcrumbs.

tbm:
v2: fixed a pylint error and changed the name to `pretty_breadcrumbs`

v8/pretty_breadcrumbs/.gitignore
```diff
@@ -0,0 +1 @@
+__pycache__
```
Kwpolska: should not be necessary, we have that in a global .gitignore for this repo

v8/pretty_breadcrumbs/README.md
```diff
@@ -0,0 +1,16 @@
...
+breadcrumbs based on metadata information defined in the source
+file.
+
+For example, if file `test/index.rst` contains
+
+```
+.. crumb: This is a test
+```
+
+the breadcrumb generated would be `This is a test` instead of `test`.
```
Kwpolska: how about defaulting to titles? Try `crumb` first, if it doesn’t exist, use `title`

v8/pretty_breadcrumbs/conf.py.sample
```diff
@@ -0,0 +1,5 @@
+
+# By default, the pretty_crumbs plugin looks for the metadata tag `crumb`.
+# This can be changed with PRETTY_BREADCRUMBS_TAG.
+#PRETTY_BREADCRUMBS_TAG = "nav"
```
Kwpolska: I honestly wouldn’t bother with configurability.

v8/pretty_breadcrumbs/pretty_breadcrumbs.py
```diff
@@ -0,0 +1,70 @@
...
+    def _set_pretty_crumbs(self, site):
+        old_get_crumbs = utils.get_crumbs
+
+        if 'PRETTY_BREADCRUMBS_TAG' in self.site.config:
+            tag = self.site.config['PRETTY_BREADCRUMBS_TAG']
+        else:
+            tag = 'crumb'
+
+        def pretty_get_crumbs(path, is_file=False, index_folder=None, lang=None):
+            lang = lang if lang else self.site.default_lang
```
Kwpolska: use LocaleBorg’s current language in the default case (tons of examples in Nikola core)

v8/pretty_breadcrumbs/pretty_breadcrumbs.py
```diff
@@ -0,0 +1,70 @@
...
+            crumbs = old_get_crumbs(path, is_file=is_file, index_folder=index_folder, lang=lang)
+            _crumbs = []
+            for link, text in crumbs:
+                if link == '#':
+                    file = path
+                else:
+                    file = os.path.normpath(os.path.join(path, link))
+                if not is_file:
+                    file += "/" + self.site.config['INDEX_FILE']
```
Kwpolska: Use `os.sep` instead of `"/"` for Windows compatibility

v8/pretty_breadcrumbs/.gitignore
tbm: Ok

v8/pretty_breadcrumbs/README.md
tbm: Titles are usually quite long and I think the filepath (slug) is better than the title if no `crumb` is defined.

But maybe this should be a config option.

v8/pretty_breadcrumbs/conf.py.sample
tbm: I put the config option in because I use `nav` for historical reasons but it's easy enough to change it to `crumb`. I can drop the config if you don't think it's useful.

v8/pretty_breadcrumbs/pretty_breadcrumbs.py
tbm: Ok

tbm:
> We could also think about adding this behavior to core’s utils.get_crumbs?

Well, that's up to you. Let me know if you prefer a plugin or a patch against utils.get_crumbs

tbm:
I pushed a new version. I haven't removed PRETTY_BREADCRUMBS_TAG yet but I'm happy to do that. I'll defer to you on this.

Kwpolska:
It’s a plugin, so anything goes. As for an upstream patch, I’m not really sure about it. It can live as a plugin.

PR #269, opened 2018-04-20
use <s> tag instead of deprecated <strike>

gour:
Not a big thing, but it looks as `<strike>` tag is deprecated and obsolete, so new sites are recommended to use `<s>` tag instead...


ralsina:
Could you change it so when someone uses a strike role it generates s tags?

That way you don't break current users.

PR #270, opened 2018-04-24
Added Babeldates plugin

sukiletxe:
This PR adds a plugin which formats dates using Babel (see getnikola/nikola#2606). I know the bug reported there is not on your side, but just in case someone finds it useful, I share the plugin mentioned.


sukiletxe:
Woops, tests fail.

ralsina:
Don't worry, I will fix them.

ralsina:
Because I have not fixed them :-)

felixfontein:
I guessed so ;) Just wanted to ping this in case you forgot ;)

ralsina:
42d6e2b

sukiletxe:
Thanks. I'm a disaster with wrapping.

ralsina:
NP, besides the real problem is that the thing we do to merge requirements from all plugins for testing is very stupid.

PR #275, opened 2018-05-10
static_comments: call mako filters inside a function

evgeni:
you can't call mako filters inside python code, you'll get an error like
this instead:

        '<span class="author">' + ('<a href="{0}">{1}</a>'.format(comment.author_url|h, comment.author|h) if comment.author_url is not None else (comment.author|h)) + '</span>',
    TypeError: unsupported operand type(s) for |: 'str' and 'Undefined'

move the author line generation into an mako function, which again can
use mako filters

### DIFF
--- a/v7/static_comments/templates/mako/static_comments_helper.tmpl
+++ b/v7/static_comments/templates/mako/static_comments_helper.tmpl
@@ -12,8 +12,15 @@
         <div class="comment-header">
           <a name="comment-${ comment.id }"></a>
           % if comment.author is not None:
+            <%def name="author()" buffered="True">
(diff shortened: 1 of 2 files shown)


evgeni:
@felixfontein now the comments render with mako too ;)

felixfontein:
Meh, I really prefer jinja2 :D

felixfontein:
Again, thanks for fixing this! :)

evgeni:
@felixfontein heh, yeah. Today I'd use j2 too. But my blog is still using mako and I did not find the time to migrate the custom templates.

PR #298, opened 2019-02-14
A plugin to hide the input cells

SuryaSankar:


ralsina:
Isn't this the same as adding that CSS in conf.py?

SuryaSankar:
I wasn't aware that there was a method to add CSS in conf.py.  The documentation indicated that the way to insert css snippets common across all themes was to use site.template_hoooks['extra_head']. So tried this approach, so that the same will work on all themes.

I am trying to create multiple sites using Nikola. Will be nice if I can just enable this functionality using `nikola plugin -i hide_input_cells` instead of having to edit a file in every project.

v8/hide_input_cells/hide_input_cells.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import unicode_literals
```
Kwpolska: We recommend adding a license snippet to all plugins.

v8/hide_input_cells/hide_input_cells.py
```diff
@@ -0,0 +1,20 @@
...
+from nikola.plugin_categories import ConfigPlugin
+
+
+class HideInputCells(ConfigPlugin):
+    def set_site(self, site):
+        site.template_hooks['extra_head'].append(
+            """
+            <!--Code inserted by HideInputCells -->
+            <style type="text/css">
+              div.input {
```
Kwpolska: Those selectors feel too wide, are you sure you can’t use something that would only work on Jupyter notebooks and not all potential `div.input`s and `div.prompt`s out there?

PR #299, opened 2019-02-19
sidebar: exclude drafts, private and timed posts

mardy:


felixfontein:
@mardy thanks for the PR!
@Kwpolska thanks for merging!

PR #300, opened 2019-02-28
make `jsonfeed` work in v8

tedder:
I don't know if I am supposed to copy this to `v8/`, but these are the mimimum changes required for jsonfeed to not break a v8 installation:
- removed unused config entries that have been deprecated/renamed
- add a default for the one config entry that is actually used

I don't know if it makes jsonfeed **work**, I just know that it doesn't prevent nikola from starting.

### DIFF
--- a/v7/jsonfeed/jsonfeed.py
+++ b/v7/jsonfeed/jsonfeed.py
@@ -62,7 +62,7 @@ class JSONFeed(Task):
             'feed_links_append_query': self.site.config['FEED_LINKS_APPEND_QUERY'],
             'feed_length': self.site.config['FEED_LENGTH'],
             'feed_plain': self.site.config['FEED_PLAIN'],
-            'feed_previewimage': self.site.config['FEED_PREVIEWIMAGE'],
+            'feed_previewimage': self.site.config.get('FEED_PREVIEWIMAGE', True),
             'feed_read_more_link': self.site.config['FEED_READ_MORE_LINK'],
             'feed_teasers': self.site.config['FEED_TEASERS'],
             'jsonfeed_append_links': self.site.config.get('JSONFEED_APPEND_LINKS', True),
@@ -70,9 +70,6 @@ class JSONFeed(Task):
             'blog_title': self.site.config['BLOG_TITLE'],
             'blog_description': self.site.config['BLOG_DESCRIPTION'],
             'blog_author': self.site.config['BLOG_AUTHOR'],
-            'tag_pages_titles': self.site.config['TAG_PAGES_TITLES'],
-            'category_pages_titles': self.site.config['CATEGORY_PAGES_TITLES'],
-            'posts_section_title': self.site.config['POSTS_SECTION_TITLE'],
             'archives_are_indexes': self.site.config['ARCHIVES_ARE_INDEXES'],
         }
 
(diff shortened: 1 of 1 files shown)


felixfontein:
The general idea is (I think): if the change will not break the module for v7, you can keep the plugin in `v7/`. If for v8 you need to do changes which are incompatible to v7, copy it to `v8/` and do the changes there.

v7/jsonfeed/jsonfeed.py
```diff
@@ -62,17 +62,14 @@ def set_site(self, site):
...
-            'feed_previewimage': self.site.config['FEED_PREVIEWIMAGE'],
+            'feed_previewimage': self.site.config.get('FEED_PREVIEWIMAGE', True),
             'feed_read_more_link': self.site.config['FEED_READ_MORE_LINK'],
             'feed_teasers': self.site.config['FEED_TEASERS'],
             'jsonfeed_append_links': self.site.config.get('JSONFEED_APPEND_LINKS', True),
             'site_url': self.site.config['SITE_URL'],
             'blog_title': self.site.config['BLOG_TITLE'],
             'blog_description': self.site.config['BLOG_DESCRIPTION'],
             'blog_author': self.site.config['BLOG_AUTHOR'],
-            'tag_pages_titles': self.site.config['TAG_PAGES_TITLES'],
```
Kwpolska: You could make it work with both versions properly by checking if `TAG_PAGES_TITLES` exists and adding the three entries based on that.

v7/jsonfeed/jsonfeed.py
```diff
@@ -102,7 +99,8 @@ def gen_tasks(self):
                                           timeline, feed_url, output_name)
 
             for classification_name, path_handler in self.supported_taxonomies.items():
-                taxonomy = self.site.taxonomy_plugins[classification_name]
+                taxonomy = self.site.taxonomy_plugins.get(classification_name)
+                if not taxonomy: continue
```
Kwpolska: How was this failing? (Also, please write this in two lines.)

v7/jsonfeed/jsonfeed.py
tedder: but AFAICT this key isn't used anyhow.

v7/jsonfeed/jsonfeed.py
tedder: It's failing like this:
```
  File "mysite/plugins/jsonfeed/jsonfeed.py", line 102, in gen_tasks
    taxonomy = self.site.taxonomy_plugins[classification_name]
KeyError: 'section_index'
````

I may be in over my head on trying to make quick little changes to get this working. If so it's probably worth just killing this PR.

v7/jsonfeed/jsonfeed.py
Kwpolska: It’s not used directly, but values from the dict will eventually end up here:

https://github.com/getnikola/plugins/blob/765888e315d66180ef79428ab2d01aa15245abae/v7/jsonfeed/jsonfeed.py#L121-L122

`self.kw` is also used to ensure stuff is rebuilt when a config setting is changed.

Kwpolska:
Superseded by commit d3ce6e1. Thanks for spotting that this plugin was broken.

PR #312, opened 2019-09-22
Accordion Directive

hx2A:
Accordion Directive using Bootstrap.

### DIFF
--- a/v8/accordion/README.md
+++ b/v8/accordion/README.md
@@ -1,0 +1,53 @@
+Accordion for reStructuredText.
+
+Separate each box from the next with two blank lines. The first line in each group will become the title. All of the remaining lines in the group will become the contents.
+
+You can put arbitrary reStructuredText directives inside the accordion, including videos and images.
+
+```
+
+    .. accordion::
+
+      Box 1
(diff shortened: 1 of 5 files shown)


ralsina:
Not to be picky, we could merge it as-is, BUT

* If you implement this as a shortcode we can use it in markdown and other similar things
* You can then do the directive on top of the shortcode

There are examples in the nikola codebase if you want to go that way.

Also: Saying "no, please merge as-is" is a perfectly acceptable answer ;-)

hx2A:
I don't know how to do that. When I coded the Accordion directive I modeled it after the Slides directive. Can you suggest a different plugin for me to look at to understand how to code this as a shortcode?

And would this even work in something other than reST? Note `accordion.py` has a call to `rest.rst2html`, which is essential for someone to create an accordion with images or videos inside.

ralsina:
I see. Oh well, let's merge it as-is then.

Thanks for the contribution!

hx2A:
You are welcome! This is my first Nikola contribution.

hx2A:
When will this be available for users to install? I just tried using the command `nikola plugin -i accordion` and it said it couldn't find the plugin. Also see it isn't mentioned on https://plugins.getnikola.com/.

Kwpolska:
> When will this be available for users to install? I just tried using the command nikola plugin -i accordion and it said it couldn't find the plugin. Also see it isn't mentioned on https://plugins.getnikola.com/.

The site is meant to be updated automatically every day; this did not happen due to a misconfiguration. I fixed the scripts and rebuilt the site.

> Also, I just noticed that this directive requires bootstrap4 to work correctly. It does not work with bootstrap3. It is an easy fix and I will take care of it.

If you can make it compatible with both, then that’s great. Otherwise, we don’t need to support the bootstrap3 theme.

> How do I go about doing that with this pull request? Do I make a new pull request? Do I need to update my fork?

You need to make a new pull request. You will need to update your fork:

    git remote add upstream git@github.com:getnikola/plugins.git
    git pull upstream master

After that, you can make new commits, push to the `origin` remote, and make a new PR.

hx2A:
I maintain a nikola website that uses bootstrap3 so I'll try to make it compatible.

Is there a way for the directive to tell if the user's theme uses bootstrap3 or bootstrap4? A solution will require the directive to pick one of two templates.

Kwpolska:
While there are some possibilities with using the theme inheritance tree,
this doesn't work for custom themes that don't inherit from our bootstrap3
theme.

-- 
Chris Warrick <https://chriswarrick.com/>

hx2A:
I maintain a small website that uses bootstrap3 and it was easier to fix the directive than change the custom theme. It was an easy fix. I also made some other minor improvements.

In the next week or so I'll make Jinja templates, test, and make a new pull request. Thanks for the advice.

PR #313, opened 2019-10-03
improve accordion directive to support bootstrap3 and jinja themes

hx2A:
As discussed in another thread, I have improved the Accordion directive to support bootstrap3 themed sites.

In addition, I fixed a bug with jinja2 themes and made some other minor improvements.

While testing this I also discovered a bug in the slides directive. I'll make a separate pull request for that fix.


v8/accordion/accordion.py
```diff
@@ -30,7 +32,18 @@ def set_site(self, site):
...
+        super().__init__(name, arguments, options, content, lineno,
+                         content_offset, block_text, state, state_machine)
+        self.state_id = id(state)
+
     has_content = True
+    optional_arguments = 1
+    # the purpose of this counter stuff is to give each Accordion on a page a
+    # unique number starting with one. These numbers become a part of the
+    # Accordion's html id and aria controls via the template.
+    counters = defaultdict(count)
```
Kwpolska: This won’t work, please restore the UUIDs. You will have duplicate numbers if two different posts with accordions are on one page (with indexes) and they are built in separate `nikola build` runs.

hx2A: Understood. That's a good point, I didn't realize I was introducing a problem by changing that. Thank you for the feedback.

I just restored the UUIDs.

Kwpolska:
Thanks for contributing!

hx2A:
You are welcome! Thank you for helping me become a better contributor. I have another idea for a new directive that I'll start working on in the near future.

PR #316, opened 2019-10-19
add command when org-mode failed

huhuang03:
When org-mode failed, I have to manually concat the command from the log.

It's maybe better print out the run field command automatic.


v8/orgmode/orgmode.py
```diff
@@ -90,7 +90,8 @@ def compile(self, source, dest, is_two_file=True, post=None, lang=None):
                 req_missing(['emacs', 'org-mode'],
                             'use the orgmode compiler', python=False)
         except subprocess.CalledProcessError as e:
-            raise Exception('Cannot compile {0} -- bad org-mode configuration (return code {1})'.format(source, e.returncode))
+            raise Exception('''Cannot compile {0} -- bad org-mode configuration (return code {1})
+The command: is {2}'''.format(source, e.returncode, " ".join(e.cmd)))
```
Kwpolska: Use `shlex.join` so that commands with spaces work correctly when copied. Also, please move the colon after `is`.

huhuang03: The `shlex.join` is only exist in `python3.8`. And not exist even in `python3.7`. Any other good replacement?

Kwpolska: shlex.join is a one-liner, you can just do this:

```py
' '.join(shlex.quote(arg) for arg in e.cmd)
```

huhuang03: Oh, clear.

Kwpolska:
Thanks!

PR #322, opened 2020-01-10
Add org-mode support for internal links (incl. update documentation)

dalanicolai:
Similar to the problem with image url's, the url's of internal link are also not handled well by ox-html for correct use with nikola. I copied and slightly the solution for handling image urls so that internal urls will also be handled correctly now with org-mode.

In addition to changing the implementing the code I updated the documentation in the README file for the plugin.

### DIFF
--- a/v8/orgmode/README.md
+++ b/v8/orgmode/README.md
@@ -44,3 +44,7 @@ because the path is considered as an absolute file path.
 
 In order to correctly generate image urls, you may write `[[img-url:/images/test.jpg]]`,
 and then it should be generated as `<img src="/images/test.jpg" alt="test.jpg">`.
+
+The internal link in ox-html is a little fuzzy. For example, `[[./test.org][test]]` will be generated as `<a href="file:///./test.jpg">"test"` because the path is considered as an absolute file path.
+
+In order to correctly generate internal link urls, you may write `[[int-url:./test][test]` (i.e. JUST THE FILENAME WITHOUT THE .org EXTENSION PART), and then it should be generated as <a href="./test/index.html">"test".
--- a/v8/orgmode/init.el
+++ b/v8/orgmode/init.el
@@ -111,6 +111,13 @@ contextual information."
             (pygmentize (downcase lang) (org-html-decode-plain-text code)))
         code-html))))
 
+;; Export internal links with custom link type
+(defun org-custom-internal-link-url-export (path desc format)
+  (cond
+   ((eq format 'html)
+    (format "<a href=\"%s/index.html\">\"%s\"" path desc))))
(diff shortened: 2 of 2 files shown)


v8/orgmode/README.md
```diff
@@ -44,3 +44,7 @@ because the path is considered as an absolute file path.
 
 In order to correctly generate image urls, you may write `[[img-url:/images/test.jpg]]`,
 and then it should be generated as `<img src="/images/test.jpg" alt="test.jpg">`.
+
+The internal link in ox-html is a little fuzzy. For example, `[[./test.org][test]]` will be generated as `<a href="file:///./test.jpg">"test"` because the path is considered as an absolute file path.
```
Kwpolska: I don’t think you meant `.jpg` here.

v8/orgmode/init.el
```diff
@@ -111,6 +111,13 @@ contextual information."
             (pygmentize (downcase lang) (org-html-decode-plain-text code)))
         code-html))))
 
+;; Export internal links with custom link type
+(defun org-custom-internal-link-url-export (path desc format)
+  (cond
+   ((eq format 'html)
+    (format "<a href=\"%s/index.html\">\"%s\"" path desc))))
```
Kwpolska: Don’t include `index.html` in those links. It’s not necessary, and we allow changing the filename.

v8/orgmode/README.md
dalanicolai: Indeed forgot to change that one... I was tired after studying how it all worked

v8/orgmode/init.el
dalanicolai: Ah thanks didn't know that. just starting with nikola

v8/orgmode/README.md
```diff
@@ -44,3 +44,7 @@ because the path is considered as an absolute file path.
...
+
+In order to correctly generate internal link urls, you may write `[[int-url:./test][test]` (i.e. JUST THE FILENAME WITHOUT THE .org EXTENSION PART), and then it should be generated as <a href="./test/index.html">"test".
```
punchagan: I'm probably bike shedding here, but `[[internal:./test][test]` seems better than `[[int-url:./test][test]` to me. I'll leave it up to your judgement.

dalanicolai: Yes good suggestion. To me the internal: looks better also.

dalanicolai:
I have implemented your suggestions/requirements for accepting the pull request

Kwpolska:
I don’t know anything about org-mode, but could you show a more specific case where this style of linking would actually be useful? Can’t you just use links like 'foo.jpg' and get the correct output? And as for the part with .org specifically, why would you want to use that?

PR #323, opened 2020-02-10
Update mediawiki.py

euphi:
Argument `with_dependencies` deleted as it is no longer valid.  --> Fixes broken mediawiki plugin for v8.


euphi:
Is `with_dependencies` still available for v7?

Kwpolska:
[Yes, it is.](https://github.com/getnikola/nikola/blob/v7.8.15/nikola/nikola.py#L1737-L1738)

euphi:
Ok, done - change is aplied to v8 only.

PR #324, opened 2020-02-17
Removed space after SIDEBAR_MAXIMUM_POST_COUNT

o1knives:
An extra space was causing sidebar to ignore SIDEBAR_MAXIMUM_POST_COUNT in conf.py.

### DIFF
--- a/v7/sidebar/sidebar.py
+++ b/v7/sidebar/sidebar.py
@@ -130,7 +130,7 @@ class RenderSidebar(Task):
         context = {}
         deps_dict = {}
 
-        posts = self._build_post_list(lang, self.site.config.get("SIDEBAR_MAXIMUM_POST_COUNT ", 10))
+        posts = self._build_post_list(lang, self.site.config.get("SIDEBAR_MAXIMUM_POST_COUNT", 10))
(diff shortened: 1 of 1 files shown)


o1knives:
I'm new to github, not sure why error checks are failing, I just deleted a trailing space in a string.

felixfontein:
@o1knives it looks like these CI errors are unrelated. One of them is because CI runs Python 3.4, which isn't supported by something anymore: https://travis-ci.org/getnikola/plugins/jobs/651315532?utm_medium=notification&utm_source=github_status

Another one happens during installing Nikola in CI: https://travis-ci.org/getnikola/plugins/jobs/651315536?utm_medium=notification&utm_source=github_status (same for Python 3.5 to 3.7 tests)

These are totally unrelated to your PR.

(If something would be related to your PR, it would probably be something like that some tests "depend" on this bug being in place, and they should be changed as well. But the tests don't even get far enough to fail for this reason, they stop right at the beginning.)

@Kwpolska can you take a look?

Kwpolska:
I fixed the (unrelated) tests and merged this.

PR #327, opened 2020-03-13
Improving the Accordion documentation based on user feedback

hx2A:
I added a link to the [Bootstrap Accordion documentation](https://getbootstrap.com/docs/4.0/components/collapse/#accordion-example) so readers can see what is meant by an accordion.


Kwpolska:
Thanks!

PR #331, opened 2020-04-21
plugin: devto, push new articles directly to your dev.to account

MattiooFR:
Following how the `medium`plugin was done I implemented a new plugin that can post any blog posts (written in `.md` or `.rst`) to your https://dev.to profile.

You just need to get the token as explained in the README.me file.

The plugin also check if the posts were already published to avoid republishing something that was already published.

### DIFF
--- a/v8/devto/devto_plugin.py
+++ b/v8/devto/devto_plugin.py
@@ -25,8 +25,6 @@
 # SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 
-from __future__ import print_function, unicode_literals
-
 import json
 import os
 
--- a/v8/devto/devto_plugin.py
+++ b/v8/devto/devto_plugin.py
@@ -64,12 +64,12 @@ class CommandDevto(Command):
         self.site.scan_posts()
 
         posts = self.site.timeline
-        toPost = [post for post in posts if not next((item for item in articles if item["title"] == post.title()), False) and post.meta('devto')]
+        to_post = [post for post in posts if not next((item for item in articles if item["title"] == post.title()), False) and post.meta('devto')]
 
-        if len(toPost) == 0:
+        if len(to_post) == 0:
             print("Nothing new to post...")
 
-        for post in toPost:
+        for post in to_post:
             if post.source_ext() == '.md':
                 with open(post.source_path, 'r') as file:
                     data = file.readlines()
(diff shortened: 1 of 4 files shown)


MattiooFR:
Why did it fail ?

Kwpolska:
Your code style does not match PEP 8:

```
v8/devto/devto_plugin.py:84:21: E128 continuation line under-indented for visual indent
v8/devto/devto_plugin.py:88:34: E251 unexpected spaces around keyword / parameter equals
v8/devto/devto_plugin.py:88:36: E251 unexpected spaces around keyword / parameter equals
v8/devto/devto_plugin.py:94:17: E128 continuation line under-indented for visual indent
```

Please fix those errors. We also expect PEP 8-style variable names, so please replace `toPost` with `to_post`.

MattiooFR:
Thanks for the informations, now it should be good !

v8/devto/devto_plugin.py
```diff
@@ -0,0 +1,92 @@
...
+# KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+# WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+# PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS
+# OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+# OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+# SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+from __future__ import print_function, unicode_literals
```
ralsina, line 28: Nikola is py3-only nowadays so no need for this line.

v8/devto/devto_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+            return False
+        with open('devto.json') as inf:
+            creds = json.load(inf)
+        api = pydevto.PyDevTo(api_key=creds['TOKEN'])
+
+        articles = api.articles()
+        self.site.scan_posts()
+
+        posts = self.site.timeline
+        to_post = [post for post in posts if not next((item for item in articles if item["title"] == post.title()), False) and post.meta('devto')]
```
Kwpolska, line 65: This is unnecessarily complicated. Try this:

```py
devto_titles = {item["title"] for item in articles}
to_post = [post for post in posts if post.title() not in devto_titles and post.meta('devto')]

File: v8/devto/devto_plugin.py [hunk 3]
```diff
@@ -0,0 +1,90 @@
...
+
+        if len(to_post) == 0:
+            print("Nothing new to post...")
```
Kwpolska, line 68: Perhaps you could use logging for this instead of print?

v8/devto/devto_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+
+        for post in to_post:
+            if post.source_ext() == '.md':
+                with open(post.source_path, 'r') as file:
+                    data = file.readlines()
```
Kwpolska, line 73: `file.read()` is enough, and you can rewrite this to avoid code duplication (set `data` depending on format or `continue`, then do `create_article` only once).

v8/devto/devto_plugin.py
MattiooFR: Thanks you are right, this is more simple to understand !

PR #332, opened 2020-04-21
feat: improved medium plugin, it now checks if article was already published

MattiooFR:
I used the same logic I developed for the dev.to plugin to check if the article was already published.


ralsina:
Nice! Thanks for the contribution!

PR #333, opened 2020-05-05
feat: removing the table of content and adding the title if not present before posting to medium

MattiooFR:
I improved the medium plugin by using BeautifulSoup to remove the generated table of content because it's totally messing the medium article.

I also added a `<h1>` tag with the post title in the article if it's not present already because filling the title attribute in the medium api doesnt work and the article ends up being without any title.

I also formatted the previous code to be compliant with PEP style.

### DIFF
--- a/v8/medium/medium_plugin.py
+++ b/v8/medium/medium_plugin.py
@@ -32,6 +32,7 @@ import os
 from medium import Client
 from nikola import utils
 from nikola.plugin_categories import Command
+from lxml import html, etree
 
 from bs4 import BeautifulSoup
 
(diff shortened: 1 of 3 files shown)


v8/medium/medium_plugin.py
```diff
@@ -33,7 +33,9 @@
 from nikola import utils
 from nikola.plugin_categories import Command
 
-LOGGER = utils.get_logger('Medium')
+from bs4 import BeautifulSoup
```
Kwpolska, line 36: Please use `lxml.html` instead. Nikola uses lxml for all its HTML parsing needs, and bs4 would be an extra unnecessary dependency.

Kwpolska:
From [Medium’s API docs](https://github.com/Medium/medium-api-docs#33-posts):

> The title of the post. Note that this title is used for SEO and when rendering the post as a listing, but will not appear in the actual post—for that, the title must be specified in the content field as well. 

I knew Medium was dumb, but never that it was completely braindead.  Oh well.

v8/medium/medium_plugin.py
MattiooFR: Ok that will be my goal for tomorrow ahah I should have checked before going straight for BeautifulSoup

MattiooFR:
My code works for the TOC made with pandoc, but I don't know how rst is generating the table of content in html so I don't know if it will works for removing the toc.

Kwpolska:
It uses div id="contents", but I would be wary about removing it, since that ID could happen in organic content.

MattiooFR:
Ok, just thought that I forgot to remove BeautifulSoup from the requirement.txt 🤦🏻‍♂️

MattiooFR:
Mmh, it is not in the requirements.txt anymore, maybe I actually did remove it ^^nevermind

MattiooFR:
Oh no even better, I actually forgot to put it there the first time ahah well at least everything is fine. Have a great day!

PR #335, opened 2020-05-10
[asciidoc] Repair user-provided ASCIIDOC_OPTIONS

jpgrayson:
The ASCIIDOC_OPTIONS value from conf.py was being inserted at the end of
the asciidoc command line, after the input file ('-'). This causes
asciidoc to fail with:

    asciidoc: Too many arguments

The solution is to insert ASCIIDOC_OPTIONS before the input file
argument.

Signed-off-by: Peter Grayson <pete@jpgrayson.net>


ralsina:
LGTM thanks for the patch!

Issue #334, opened 2020-05-09
[orgmode] Enable to export inline images with file scheme


shishimaru:
[ Background ]
Currently we need to refer inline images with dedicated scheme img-url.
However, during editing with orgmode, Emacs orgmode can't recogine the scheme and it is inconvenient because they are invisible on Emacs.

[ Task ]
Support file scheme supported by Emacs orgmode.

Kwpolska:
Are you willing to provide a patch? (Nikola is not going to support `/foo.png` or `foo.png` in the way orgmode expects; if you want to make a `file://` scheme, and it can be used with relative URLs, we can accept that patch.)

shishimaru:
@Kwpolska 
Yes, I have the patch. Could I push my working branch ?

Kwpolska:
@shishimaru Make a pull request.

shishimaru:
@Kwpolska 
I made the pull request. Could you check ?

PR #336, opened 2020-05-13
feat(orgmode): Enabled to export inline images with file scheme. Refs…

shishimaru:
… #334


Kwpolska:
Please add documentation to README.md that explains how this is meant to be used.

shishimaru:
@Kwpolska 
I added the description. Could you check ?

PR #338, opened 2020-05-18
Update continuous_import.py

jjconti:
Fix module docstring.


jjconti:
or maybe "Import and merge feeds into your blog." which is already twice in the file.

PR #344, opened 2020-06-15
ISSUE-343: Plantoid and Python3

tzulberti:
Remove python3 restriction when using PeeWee

### DIFF
--- a/v7/planetoid/planetoid.py
+++ b/v7/planetoid/planetoid.py
@@ -30,7 +30,6 @@ import datetime
 import hashlib
 from optparse import OptionParser
 import os
-import sys
 
 from doit.tools import timeout
 from nikola.plugin_categories import Command, Task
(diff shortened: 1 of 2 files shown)


v7/planetoid/planetoid.py
```diff
@@ -76,12 +76,9 @@ def init_db(self):
...
-        if peewee is None or sys.version_info[0] == 3:
-            if sys.version_info[0] == 3:
-                message = 'Peewee, a requirement of the "planetoid" command, is currently incompatible with Python 3.'
-            else:
-                req_missing('peewee', 'use the "planetoid" command')
-                message = ''
+        if peewee is None:
+            req_missing('peewee', 'use the "planetoid" command')
+            message = ''
             yield {
```
Kwpolska, line 82: This `yield` block and the `message` variable can be removed as well.

Kwpolska:
Thanks for fixing this!

PR #349, opened 2020-08-27
Send markdown instead of HTML

MattiooFR:
I fixed a bug where the HTML sent was including link to a non working URL in every line of the medium article. I found that by simply sending the markdown, medium was handling it perfectly. I kept the mechanism of adding a title first if there wasn't. 
I also added a sentence that says : "Original article : https://urltoarticle.com" below the title

**Known issue to fix :** 
When sending code blocks inside numbered list medium is reseting the number, I will be trying to replace the numbered list to a bulleted list in an attempt to make it look good in the coming days.

### DIFF
--- a/v8/medium/medium_plugin.py
+++ b/v8/medium/medium_plugin.py
@@ -28,11 +28,11 @@ from __future__ import print_function, unicode_literals
 
 import json
 import os
+import re
 
 from medium import Client
 from nikola import utils
 from nikola.plugin_categories import Command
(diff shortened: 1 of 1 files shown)


ralsina:
But there's no guarantee that posts in Nikola are markdown.

MattiooFR:
You make a point, indeed 😆 
I will try to find another way then

Kwpolska:
Closing due to no activity.

PR #353, opened 2020-09-24
postcast: allow setting a different base url for enclosure files

### LINKED ISSUE #352: [postcast] allow separate handling of enclosure files for podcasts to support hosting on CDN
Hi,

I'd like to host my podcast files at a completely different location (let's say on a CDN) than my blog.
For that I'd like to have the podcast files _somewhere_ in my file system, but not be part of the build output, but still reference them from my blog posts. I think this can be done by adding an (optional) extra config var for the plugin where you could set the CDN URL, and an another (optional) extra config var where you could set the local enclosure folder.

tikki:
adds two new config vars `POSTCAST_BASE_URL` and `POSTCAST_ENCLOSURE_FOLDER` which allow to use a base URL completely different from your blog's URL, and to keep the podcast files locally in a separate location, so they don't have to be part of the build process and can be synced separately.

fixes #352

### DIFF
--- a/v8/postcast/postcast.py
+++ b/v8/postcast/postcast.py
@@ -178,7 +178,7 @@ class Postcast (Task):
 
     def audio_url(self, lang=None, post=None):
         config = self.site.config
-        baseurl = config.get('POSTCAST_BASE_URL') or config['BASE_URL']
+        baseurl = config.get('POSTCAST_BASE_URL', config['BASE_URL'])
         return urljoin(baseurl, self.audio_path(lang=lang, post=post, is_link=True))
 
     def audio_path(self, lang=None, post=None, is_link=False):
--- a/v8/postcast/postcast.py
+++ b/v8/postcast/postcast.py
@@ -178,14 +178,20 @@ class Postcast (Task):
 
     def audio_url(self, lang=None, post=None):
         config = self.site.config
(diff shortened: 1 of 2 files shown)


v8/postcast/postcast.py
```diff
@@ -178,14 +178,20 @@ def enclosure(self, post=None, lang=None):
 
     def audio_url(self, lang=None, post=None):
         config = self.site.config
-        return urljoin(config['BASE_URL'], self.audio_path(lang=lang, post=post, is_link=True))
+        baseurl = config.get('POSTCAST_BASE_URL') or config['BASE_URL']
```
Kwpolska, line 181: You can simplify this with: `config.get('POSTCAST_BASE_URL', config['BASE_URL'])`

You can also do the same below.

tikki: Thanks for the quick response!

> You can simplify this with: `config.get('POSTCAST_BASE_URL', config['BASE_URL'])`

It's not strictly the same, if `POSTCAST_BASE_URL` was set to an empty string this would still default to the `BASE_URL`, but it's fine with me either way.

> You can also do the same below.

In the code below `POSTCAST_ENCLOSURE_FOLDER` is not simply an alternative data source, if it's present it changes the behavior of how the path is constructed, I don't see how that can be done with a default for `config.get`.

PR #359, opened 2020-11-03
Add the Quotebacks plugin

devilgate:
This is a plugin to support using the Quotebacks JavaScript library via Markdown (as discussed [here](https://github.com/getnikola/nikola/pull/3460)).

### DIFF
--- a/v8/quotebacks/quotebacks_mdx.py
+++ b/v8/quotebacks/quotebacks_mdx.py
@@ -1,0 +1,240 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+# Quotebacks extension for Python-Markdown
+
+Use by including `QuotebacksExtension()` in the extensions list.
+
+Converts Markdown of the form:
+
+```markdown
+> QUOTED-CONTENT
+>
+> -- AUTHOR, [LINKTEXT](URL)
+```
+
+to the quotebacks format.
+"""
+
+
+import logging
+import markdown
+import re
+import sys
+import xml.etree.ElementTree as ET
+
+from markdown import Extension
+
+from nikola.plugin_categories import MarkdownExtension
+
+LOGGER = logging.getLogger("quotebacks-mdx")
+
+# If required, add this at the bottom of the page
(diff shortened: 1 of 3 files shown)


v8/quotebacks/quotebacks_mdx.py
```diff
@@ -0,0 +1,240 @@
...
+
+        Keys set in self.config can be passed in as kwargs to this
+        extension to override the default values. These are available
+        from self.getConfigs() (as a dict) later.
+        """
+        # self.config["key"] = [
+        #    [],  # default value
+        #    "Sets ...",   # description of parameter
+        # ]
+        self.config = {}
```
Kwpolska, line 55: Is this required? What can users put here? Is this relevant to a Nikola user? If yes, it should be done via `conf.py`.

v8/quotebacks/quotebacks_mdx.py
```diff
@@ -0,0 +1,240 @@
...
+                "cite": cite_url,
+            }
+
+            [bq.set(k, v) for k, v in attrib.items()]
+
+            LOGGER.info("Successful: blockquote -> quoteback")
+            self.md.quotebacks_found = True
+
+
+def test():
```
Kwpolska, line 177: This isn’t relevant to Nikola, perhaps remove it.

v8/quotebacks/quotebacks_mdx.py
```diff
@@ -0,0 +1,240 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
```
Kwpolska, line 2: Since you’re using third-party code, make sure to include a license header and a license file in the plugin directory. Mentioning the license in README would be a good idea as well.

Kwpolska (changes requested):

v8/quotebacks/quotebacks_mdx.py
devilgate: Removed, after many months.

v8/quotebacks/quotebacks_mdx.py
devilgate: Done.

v8/quotebacks/quotebacks_mdx.py
devilgate: Done

PR #362, opened 2020-12-19
Fix some test warnings that are spamming the log

matthew16550:


Kwpolska:
Could you perhaps rewrite the suite into pytest? You can use the Nikola test suite for inspiration, and if you don’t feel like it, this one can be merged too.

matthew16550:
Hi Chris, I don't have the head space right now for rewriting as focused on the PlantUML plugin.  This just fixes some spam that is distracting me!

Kwpolska:
I merged this, thanks for fixing this anyway!

PR #363, opened 2020-12-19
Add new v8/sass

rensite:
I have added new sass plugin which is using python's [libsass](https://github.com/sass/libsass-python) package.
The main difference from `v7/sass` is that libsass wrapper over the sass extension installs along with the plugin.


v8/sass/sass.plugin
```diff
@@ -0,0 +1,13 @@
+[Core]
+Name = sass
+Module = sass
+
+[Nikola]
+PluginCategory = Task
+MinVersion = 7.0.0
```
Kwpolska, line 7: This isn’t necessary.

Kwpolska (changes requested):
The old plugin could support multiple *.sass/*.scss files per theme. Since this plugin is replacing it, it should offer the same feature set. Please add all the features and API of the old plugin, otherwise it will not be accepted.

PR #364, opened 2020-12-29
Add "plantuml" plugin

matthew16550:
Cleanest to flatten this when merging as there are several commits.


matthew16550:
I also have a `plantuml-markdown` plugin mostly ready, it will use the rendering functionality from here.

v8/plantuml/README.md
```diff
@@ -0,0 +1,30 @@
...
+# Known Issues
+
+- It's slow!  Every PlantUML rendering launches a new Java process, on my laptop it takes 4-8 seconds per file.
+  I have some ideas to speed this up, and they may be available in future plugin versions.
+
+- Changes to files included via `!include ...` or via a pattern (e.g. `-Ipath/to/*.iuml`) will NOT trigger a rebuild.
+  Instead, if you include them explicitly in `PLANTUML_ARGS` (e.g. `-Ipath/to/foo.iuml`) then they will trigger a
+  rebuild.
+
+- `nikola auto` does not watch dirs in `PLANTUML_FILES` or files included via `PLANTUML_ARGS` / `!include`.
```
matthew16550, line 20: Perhaps in a separate PR we could add an`EXTRA_WATCHED_FOLDERS` config option or `BasePlugin.register_watched_folder()` function to Nikola?

Kwpolska: :+1: from me for `register_auto_watched_folder`, the config option is not necessary.

v8/plantuml/README.md
```diff
@@ -0,0 +1,30 @@
...
+Developed against PlantUML version 1.2020.24.  Probably works with some earlier versions.
+
+# Unicode
+
+The plugin expects PlantUML files to be encoded with UTF-8.
+
+# Known Issues
+
+- It's slow!  Every PlantUML rendering launches a new Java process, on my laptop it takes 4-8 seconds per file.
+  I have some ideas to speed this up, and they may be available in future plugin versions.
```
Kwpolska, line 14: What are those ideas? Mine would be a PlantUML compile server that starts up with the first PlantUML file encountered, and that is reused for the rest — but note that this might be tricky with parallel builds (`nikola build -n 2`). (You are free to implement whatever you want.)

matthew16550: I have it working with [JPype](https://jpype.readthedocs.io/en/latest/) in a separate branch but I'd like to add it later as a separate PR, this one feels big enough already.

Was amazingly simple to do because JPype just worked for me.  From the Python perspective, rendering PlantUML is a simple function call into the Java.

It wont be the right solution for everyone so I'm planning a config option something like `PLANTUML_RUNNER` which can be set to `exec` or `jpype` to select how PlantUML is run.

The compile server was my second option but is more fiddly to code - shutting down the process / tcp connections etc.

v8/plantuml/README.md
```diff
@@ -0,0 +1,30 @@
+This plugin converts [PlantUML](https://plantuml.com/) files.
+
+The default configuration will output all `*.puml` files found under the `pages` dir as SVG files.
```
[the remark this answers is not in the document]
matthew16550: Good point.  The long term thing I'm working on is using Nikola for developer docs, in a recent work project we had quite a few sequence diagrams and after trying a few different options I found it easiest to keep the plant files in the same dir as the related markdown files.  Guess that idea has leaked over to here but you're right, the default should behave as a simpler generic website, perhaps this?

```python
PLANTUML_FILES = (
  ('plantuml/*.puml', 'plantuml', '.svg', ['-tsvg']),`
)
```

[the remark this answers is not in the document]
Kwpolska: This would be a better definition. Putting files that are not meant to be rendered as HTML in `pages` is not recommended (it works, but it’s an ugly choice).

PR #365, opened 2020-12-30
Convert test_command_tags.py & test_helloworld.py to use pytest

matthew16550:
These are the easier files to convert.  They have other aspects that could be simplified such as removing 2.x specifics but I think doing it here would over complicate the PR.

The other test files all use `ReSTExtensionTestCase` so probably belong in a separate PR.


Kwpolska:
Thanks!

PR #366, opened 2021-01-01
Fix: "publication_list" now records a dependency on ALL specified bibtex files

matthew16550:
Found while converting tests to pytest but this is clearer in its own PR.


xuhdev (approved):

xuhdev:
Thanks!

PR #367, opened 2021-01-01
Convert remaining unittests to pytest

matthew16550:
These files also use `unittest` and I am not sure if they should be touched?
* `disabled_tests/test_microdata.py`
* `tests/base.py` -  seems to be unused


Kwpolska:
`base.py` can be removed. If you get `tests_microdata` to work, you can also rewrite it to pytest.

matthew16550:
Done.

Kwpolska:
Thanks for doing this!

PR #372, opened 2021-01-10
Add "plantuml_markdown" plugin.

matthew16550:


.github/workflows/ci.yml
```diff
@@ -48,6 +48,13 @@ jobs:
           PLANTUML_EXEC: java -Djava.awt.headless=true -jar ${{github.workspace}}/plantuml.jar
         run: |
           py.test --color=yes tests/
+      - name: Archive failures
+        uses: actions/upload-artifact@v2
```
Kwpolska, line 52: I’d prefer not to, just `cat` those files in the build script, or ensure enough of them is shown by pytest.

tests/__init__.py
```diff
@@ -23,3 +35,34 @@ def execute_plugin_tasks(plugin: Task, verbosity: int = 0):
         catched = t.execute(stream)
         if catched:
             raise Exception("Task error for '{}'\n{}".format(t.name, catched.get_msg()))
+
+
+class regex:  # noqa
```
Kwpolska, line 40: This should be documented and named better. (Also, I’m not 100% sure if we need this.)

tests/conftest.py
```diff
@@ -22,23 +27,77 @@ def f(ext: str, data: str, extra_plugins_dirs: List[Path] = None, metadata: str
...
-
-        return CompileResult(tmp_site_path / 'cache' / 'pages' / 'test.html')
+        post = site.timeline[0]
+        post.compile('en')
+        return CompileResult(request, post)
 
     return f
 
 
+# CompileResult can be used as a Context Manager e.g.
```
Kwpolska, line 44: That should be a docstring in `CompileResult`.

tests/test_plantuml_markdown.py
```diff
@@ -0,0 +1,274 @@
...
+    raise pytest.skip("plantuml_markdown plugin requires Python >= 3.6", allow_module_level=True)
+
+from tests import V8_PLUGIN_PATH, regex
+from tests.conftest import CompileResult
+from v8.plantuml_markdown.plantuml_markdown import PlantUmlMarkdownProcessor, first_line_for_listing_block
+
+
+def test_svg(do_fence_test):
+    with do_fence_test('plantuml') as compiled:
+        assert compiled.raw_html == regex(
```
Kwpolska, line 18: Using regex for HTML is very unclean. Better to parse it with `lxml` and query it. (For example, what if the order of `<text>` elements becomes unstable?

v8/plantuml_markdown/README.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+```plantuml
+A -> B : This arrow uses the default style
+```
+~~~
+
+# Unicode
+
+The plugin expects Markdown files to be encoded with UTF-8.
```
Kwpolska, line 103: (So does Nikola, although we tolerate UTF-8 with BOM. You can remove this note if your plugin works with BOMs too — and if it doesn’t, try to fix that.)

v8/plantuml_markdown/README.md
```diff
@@ -0,0 +1,109 @@
+This plugin renders [PlantUML](https://plantuml.com/) in Markdown files.
+
+# Requirements
+
+* [plantuml plugin](https://plugins.getnikola.com/#plantuml) (the actual PlantUML rendering is managed there)
```
Kwpolska, line 5: Create a `requirements-plugins.txt` file and list it there (as just `plantuml`, [example](https://github.com/getnikola/plugins/blob/master/v8/pkgindex/requirements-plugins.txt))

Kwpolska (changes requested):

v8/plantuml_markdown/README.md
matthew16550: Yes BOM seems to be fine, `CompileMarkdown.compile()` reads the file and the plugin doesn't know anything about it.

tests/__init__.py
matthew16550: removed

tests/test_plantuml_markdown.py
matthew16550: You're right, xpaths make it much less brittle.

Kwpolska (approved):

PR #374, opened 2021-01-17
postcast: create extra feed pages for podcast history

### LINKED ISSUE #373: [postcast] allow podcatchers to discover all episodes
Hi,

one of the podcasts I'm hosting has reached a few dozen episodes and I noticed that by default only the latest 10 episodes are shown in most podcatchers. That is because the default feed size of Nikola is 10 and postcast simply uses that value when it creates its own feeds.
To have all podcast episodes show up I could of course increase that limit, but it would also considerably increase the feed size – both for the main posts feed and the podcast feed. This means wasted traffic and some users could even run into problems with file size limits on some setups.

A much better solution is [paged feeds](https://podlove.org/paged-feeds/), which is already supported by many podcatchers. The idea is to simply split up the feed into multiple files and [use atom links to string them all together](https://tools.ietf.org/html/rfc5005#section-3).
A nice implementation could even put only a few/the most current episodes in the base file, which is basically the entry point for the podcatchers, and then use large page sizes for the following pages to reduce the amount of requests necessary to load all the rest in case the podcatcher can't find them in their cache.

tikki:
This changes nothing from the default behaviour. Paging has to be enabled explicitly by the user.

If `POSTCAST_PAGE_LENGTH` is set, if a feed has more items than `FEED_LENGTH` this change will create further .xml files ("pages") and link them from each previous page, allowing discovery of the full archive of all podcast episodes by following those links from the original/first page.
The new files are automatically named based on the slug, by appending ".page-{page numer}", e.g. `http://example.com/feeds/podcast.xml` could have further pages at `http://example.com/feeds/podcast.page-1.xml` and `http://example.com/feeds/podcast.page-2.xml`.

fixes #373

### DIFF
--- a/v8/postcast/conf.py.sample
+++ b/v8/postcast/conf.py.sample
@@ -59,3 +59,12 @@
 #         ('Society & Culture', ('Philosophy', )),
 #     ],
 # }
+
+# Create paged feeds with the provided number of items per page.
+# The default is to not create extra feed pages.
+# The first page will always have FEED_LENGTH items.
+# Specifying a length for the '' feed will result in all unspecified
+# feeds having that length.
+# POSTCAST_PAGE_LENGTH = {
+#     '': 100,
+# }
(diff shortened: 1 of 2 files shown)


Kwpolska:
I’ll close this due to the discussion in #373 and the apparent lack of support in podcatchers.

PR #380, opened 2021-03-02
Auto register sass/less folders as watched

### LINKED ISSUE #379: Change to sass does not trigger nikola auto rebuild
This seems to be a regression of https://github.com/getnikola/nikola/issues/966 .

With the `sass` plugin installed, changing files under the `sass/` subdirectory does not trigger a rebuild when `nikola auto` is running.

Manually running `nikola build` detects the change and rebuilds the CSS from the SASS sources.

wohali:
Fixes #379 . Thanks for the advice on how to implement this, @Kwpolska !

### DIFF
--- a/v7/less/less.py
+++ b/v7/less/less.py
@@ -42,6 +42,10 @@ class BuildLess(Task):
     sources_folder = "less"
     sources_ext = ".less"
 
+    def set_site(self, site):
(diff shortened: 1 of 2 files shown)


Kwpolska:
Thanks for fixing this! (And also, sorry it took a while to get back to…)

PR #381, opened 2021-04-24
Add a plugin to use Cactus Comments

pieterdavid:
This plugin relies on the changes in https://github.com/getnikola/nikola/pull/3546 to add Cactus Comments as comment system, so it should be reviewed as part of that, or after the PR is accepted.

### DIFF
--- a/v8/cactuscomments/cactuscomments.py
+++ b/v8/cactuscomments/cactuscomments.py
@@ -28,7 +28,6 @@ class CactusComments(ConfigPlugin):
     """Use Cactus Comments https://cactus.chat/ as a comment system"""
     def set_site(self, site):
         site.template_hooks['extra_head'].append(
-                '<script type="text/javascript" src="https://latest.cactus.chat/cactus.js"></script>'
                 '<link rel="stylesheet" href="https://latest.cactus.chat/style.css" type="text/css">'
                 )
         super(CactusComments, self).set_site(site)
--- a/v8/cactuscomments/cactuscomments.py
+++ b/v8/cactuscomments/cactuscomments.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2021 Pieter David
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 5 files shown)


v8/cactuscomments/cactuscomments.py
```diff
@@ -0,0 +1,34 @@
...
+
+from __future__ import unicode_literals
+from nikola.plugin_categories import ConfigPlugin
+import nikola.utils
+
+class CactusComments(ConfigPlugin):
+    """Use Cactus Comments https://cactus.chat/ as a comment system"""
+    def set_site(self, site):
+        site.template_hooks['extra_head'].append(
+                '<script type="text/javascript" src="https://latest.cactus.chat/cactus.js"></script>'
```
Kwpolska, line 31: I think the JS script should go in the `comment_form` macro in templates,.

v8/cactuscomments/cactuscomments.py
```diff
@@ -0,0 +1,34 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from __future__ import unicode_literals
+from nikola.plugin_categories import ConfigPlugin
```
Kwpolska, line 24: We could also add a `CommentSystemPlugin` for this case (I’ll handle this someday, when both PRs are merged.)

v8/cactuscomments/cactuscomments.py
pieterdavid: Thanks! Indeed, it's only the stylesheet that needs to be here

Kwpolska (approved):

Kwpolska:
@pieterdavid I merged a bit too early, the plugin was missing a `README.md`. I added in a very simple one to unbreak the build — but please write a nice README and make a new PR.

pieterdavid:
@Kwpolska done in https://github.com/getnikola/plugins/pull/383 - thanks a lot for all the help with getting this included!

PR #385, opened 2021-05-02
Expand meta_template accepted keywords and documentation

JesperDramsch:
I recently built my website with Nikola and heavily relied on meta_templates to reuse code-bocks.

However, I needed to dig into the code to understand how to actually use it and ended up extending the code with general "template_00" keywords. I think this change (and the change to the Readme.md) could enhance the original plugin and make it more usable to the wider user base.

The main change is that the meta_template accepts "template_00" through "template_99" now. This was necessary because the options have to be pre-defined within the plugin, not accepting "any kind of option" unfortunately.

### DIFF
--- a/v7/meta_template/README.md
+++ b/v7/meta_template/README.md
@@ -32,5 +32,7 @@ You will also need to put the file `button.tmpl` in the folder
 
 Finally, you will get the button as you expect. Easy. Doesn't it?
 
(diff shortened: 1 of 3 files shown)


Kwpolska:
I think that instead of resurrecting an old plugin, it might be a better idea to use template-based [shortcodes](https://getnikola.com/handbook.html#shortcodes).

JesperDramsch:
That's fair. I didn't know about those. I think they might be less readable in some of my cases.

PR #392, opened 2021-07-02
stops: IndexError: list index out of range error

msghens:
Pluggin works as advertised. Squashes annoying IndexError.

### DIFF
--- a/v8/medium/medium_plugin.py
+++ b/v8/medium/medium_plugin.py
@@ -79,7 +79,10 @@ class CommandMedium(Command):
                 toc[0].getparent().remove(toc[0])
             if len(tree.xpath("//h1")) == 0:
                 content = "<h1>" + post.title() + "</h1>\n"
-                body = tree.xpath("//div")[0]
+                try:
+                    body = tree.xpath("//div")[0]
+                except:
+                    pass
                 body.insert(0, etree.XML(content))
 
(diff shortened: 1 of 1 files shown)


v8/medium/medium_plugin.py
```diff
@@ -79,7 +79,10 @@ def _execute(self, options, args):
                 toc[0].getparent().remove(toc[0])
             if len(tree.xpath("//h1")) == 0:
                 content = "<h1>" + post.title() + "</h1>\n"
-                body = tree.xpath("//div")[0]
+                try:
+                    body = tree.xpath("//div")[0]
+                except:
```
Kwpolska, line 84: 1. A bare `except` is not good code style, since it can also obscure other exceptions and prevent eg. `KeyboardInterrupt`s.
2. What ends up being posted if the exception is raised?
3. Is the `body` variable even used anywhere?

Kwpolska (changes requested):

v8/medium/medium_plugin.py
msghens: My initial research was trying to find the problem with lxml. I actually realized that this works even with all the IndexError warnings. 

I was more concerned with meaningless errors for myself.  Honestly, I don't even know what the body lines actually have a purpose and may just be better to be removed.

PR #398, opened 2021-09-02
Added plugin for ditaa ReST directive

spookylukey:
Includes automated tests for the happy case, mocking out the `ditaa` binary since that will be hard to install.

I have manually tested the error cases, and could add unit tests for that if wanted.

The [README](https://github.com/spookylukey/plugins/tree/ditaa-plugin/v8/ditaa) should explain the use cases. You can see it in action at https://lukeplant.me.uk/blog/posts/life-on-the-diagonal-adventures-in-2d-time/#a-2-d-calendar

### DIFF
--- a/v8/ditaa/ditaa.py
+++ b/v8/ditaa/ditaa.py
@@ -1,0 +1,121 @@
+# -*- coding: utf-8 -*-
+
+# Copyright © 2021 Luke Plant
+
+# Permission is hereby granted, free of charge, to any
+# person obtaining a copy of this software and associated
+# documentation files (the "Software"), to deal in the
+# Software without restriction, including without limitation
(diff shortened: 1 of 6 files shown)


v8/ditaa/ditaa.py
```diff
@@ -0,0 +1,95 @@
+"""Custom reST_ directive for ditaa_ integration.
```
Kwpolska, line 1: Could you add a license header, like the ones seen in Nikola core? (Pick any license you like)

spookylukey: Sure, added one.

Kwpolska (approved):

Kwpolska:
Thanks for contributing this!

spookylukey:
Thanks so much for accepting!

PR #399, opened 2021-09-03
Fixed ditaa example diagram URL

spookylukey:
With a relative URL, it works on GitHub, but with a full URL it should work on https://plugins.getnikola.com/v8/ditaa/ as well.


Kwpolska:
Thanks for fixing this!

PR #406, opened 2022-02-23
Update plugin to work in current environment

x1101:
I found that my plugin no longer worked. I've updated it to follow what appear to be more modern conventions and confirmed that it works in my current setup.


ralsina (approved):

ralsina:
Thanks for fixing this!

PR #409, opened 2022-04-23
add jquery.min.js for localsearch

damufo:
add jquery.min.js for localsearch
jquery library is required to work localsearch.
Currenly this plugin not work por example with hack theme, this fix that issue.


Kwpolska:
The default (bootstrap-based) themes already include jQuery. Adding another version can cause conflicts and unexpected behavior. This change shouldn’t be done in the plugin, since it will affect many users negatively (including users with the default theme). Instead, I added a note to the README.md file (for users of jQuery-less themes) in commit 7eca337fc908786ade2acddd0b8e687e73835d76.

PR #416, opened 2022-10-19
[publication_list] Fix: allow use of special chars in the id (from a bibtex key)

gnava:
Current code fails when a bibtex key has special characters (such as ":") when trying to set the bibtex and abstract blocks. The fix just uses the escapeSelector method to escape the ID.


xuhdev (approved):
Thank you!

PR #423, opened 2023-05-20
import_jekyll: support for image field

Mte90:
right now the new image path is kind of hardcoded because depends on the original one

### DIFF
--- a/v7/import_jekyll/import_jekyll.py
+++ b/v7/import_jekyll/import_jekyll.py
@@ -212,7 +212,15 @@ class JekyllPostImport(object):
                 return None
                 name = m.group('name')
             return name
-
+        
+        def extract_image():
+            raw_image = jmetadata.get('image')
+            new_image_path = ''
+            if isinstance(raw_image, str):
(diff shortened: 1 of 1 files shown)


v7/import_jekyll/import_jekyll.py
```diff
@@ -213,13 +213,22 @@ def extract_title():
...
+            return new_image_path
+
         tags = [x for x in jmetadata.get('tags') or [] if x]
         categories = [x for x in jmetadata.get('categories') or [] if x]
 
         metadata = PageCompiler.default_metadata.copy()
         metadata['title'] = extract_title()
         metadata['slug'] = slugify_file(path)
         metadata['date'] = extract_date()
+        metadata['previewimage'] = extract_image()
```
Kwpolska, line 231: Consider setting `previewimage` conditionally, only if an image is extracted successfully.

Mte90: I should added that check

Kwpolska (approved):

PR #425, opened 2023-05-23
Discourse Comments integration

Mte90:
Ref: https://github.com/getnikola/nikola/issues/3689

So I started working but some stuff is not clear to me.

So my needs are just 2:

* I need to check if the article has a parameter `discourse_thread` and get that value
* Print in the bottom in the comment section a HTML/JS code

So right now I saw that hooks are just 6 https://getnikola.com/extending.html#toc-entry-16 and no one of them is useful for me to print in the comment section.  
It would be useful to have something to access the macro `comment_link_script` to print what I need but I didn't find anything. I see the cactus_comment instead use template files but I am not sure if this create troubles with themes.

Instead how to read a field in the article I didn't find a plugin that do that to replicate the code.

I have some doubts if it is the case to integrate it on theme side or by plugin. Or a plugin that is not a comment system but a task generator.

PS: the code is not yet tested.

### DIFF
--- a/v8/discourse_comments/README.md
+++ b/v8/discourse_comments/README.md
@@ -1,0 +1,1 @@
+If the article includes a field `discourse_thread` it will be added on the bottom for the comments automatically the Discourse JS integration.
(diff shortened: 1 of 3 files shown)


Kwpolska:
Comment systems cannot be plugins. Instead, they should be part of Nikola core (they are just templates). Please convert your changes into Mako templates and open a pull request for getnikola/nikola.

Issue #428, opened 2023-06-25
[myst] the plugin is incompatible with myst-parser > 0.17.2


lorenzo-rovigatti:
[This PR](https://github.com/executablebooks/MyST-Parser/pull/566) restructured myst's code, breaking this plugin. A quick fix would be to have `myst-parser==0.17.2` instead of `myst-parser` in the `requirements.txt` file (and update the docs accordingly).

Kwpolska:
Could you provide a patch for this?

lorenzo-rovigatti:
I'm not exactly familiar with either myst's internals or nikola plugins, but the applying the following patch to `myst.py` seems to work:

```
--- myst_old.py 2023-06-26 19:59:23.447605537 +0200
+++ myst.py     2023-06-26 23:41:20.177490223 +0200
@@ -31,7 +31,6 @@
 
 try:
     import myst_parser
-    import myst_parser.main
 except ImportError:
     myst_parser = None
     nikola_extension = None
@@ -63,7 +62,39 @@
         if not is_two_file:
             _, data = self.split_metadata(data, post, lang)
         new_data, shortcodes = sc.extract_shortcodes(data)
-        output = myst_parser.main.to_html(new_data)
+        
+        # this works for myst-parser versions <= 0.17.2
+        try:
+            from myst_parser.main import to_html
+            output = to_html(new_data)
+        except ImportError:
+            from docutils.core import publish_string
+            from myst_parser.docutils_ import Parser
+            output = publish_string(
+                source=new_data,
+                writer_name="html5",
+                settings_overrides={
+                    "myst_enable_extensions": [ # enable all extension
+                            "amsmath",
+                            "attrs_inline",
+                            "colon_fence",
+                            "deflist",
+                            "dollarmath",
+                            "fieldlist",
+                            "html_admonition",
+                            "html_image",
+                            "linkify",
+                            "replacements",
+                            "smartquotes",
+                            "strikethrough",
+                            "substitution",
+                            "tasklist",
+                        ],
+                    "embed_stylesheet": True,
+                    'output_encoding': 'unicode',
+                },
+                parser=Parser(),
+            )
         output, shortcode_deps = self.site.apply_shortcodes_uuid(
             output, shortcodes, filename=source_path, extra_context={"post": post}
         )
```


Note that with the "new" myst-parser versions you can also enable some extensions that can come in handy (such as admonitions), but these require to style specific css classes.

lorenzo-rovigatti:
[...]

PR #429, opened 2023-07-13
Make the myst plugin compatible with myst-parser > 0.17.2

lorenzo-rovigatti:
As discussed in #428 , the myst plugin is not compatible with myst-parser > 0.17.2 (which is already more than one year old). The patch contained in this PR keeps the plugin backward compatible with older myst-parser versions, but also supports the new versions by using [docutils.core's publish_string](https://docutils.sourceforge.io/docs/api/publisher.html#publish-string), as suggested in [the docs](https://myst-parser.readthedocs.io/en/latest/docutils.html).

In principle with the `publish_string` method it is possible to list the myst extensions one wants to enable. Here I am enabling them all but support for mathematical equations (which is handled directly by nikola). I have also decided to suppress warnings related to out-of-order document headings, and to have myst add anchors to headers (i.e. lines starting with one or more #) in order to support internal links.

### DIFF
--- a/v8/myst/myst.py
+++ b/v8/myst/myst.py
@@ -79,7 +79,7 @@ class CompileMyst(PageCompiler):
                 source=new_data,
                 writer_name="html5",
                 settings_overrides={
-                    "myst_enable_extensions": 
+                    "myst_enable_extensions":
                         [
                             "attrs_inline",
                             "colon_fence",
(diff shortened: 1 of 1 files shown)


Kwpolska (approved):

Kwpolska:
Looks good; please fix the linter warnings so we can merge.

lorenzo-rovigatti:
Done (hopefully)!

PR #432, opened 2024-01-06
LaTeX formula renderer: support no unit suffix

felixfontein:
SVG files written seem to have no unit anymore. The values seem to be identical to the values before `pt` in the past.


Kwpolska (approved):

felixfontein:
@Kwpolska thanks!

PR #433, opened 2024-01-14
Added comment system plugin mastodoncomments

MasinAD:


v8/mastodoncomments/README.md
```diff
@@ -0,0 +1,40 @@
...
+        'host': 'nerdculture.de',
+        'account_id': '109270094542366763',
+        'account': 'fluchtkapsel'
+        }
+    }
+```
+See [this page](https://khendrikse.netlify.app/blog/find-your-mastodon-account-id/) on how to find your Mastodon account id.
+
+# How to install
+You need to copy the files `mastodon.js` and `mastodon.css` from `assets/files/{css,js}/` to `files/{css,js}/` respectively. Additionally, it requires DOMpurify which you can install by
```
Kwpolska, line 24: Please restructure this to use ``assets/js`` and ``assets/css`` in output, just like the rest of Nikola does.

v8/mastodoncomments/assets/files/css/mastodon.css
```diff
@@ -0,0 +1,120 @@
+.avatar {
```
Kwpolska, line 1: Consider adding `section.comments` to this selector.

v8/mastodoncomments/conf.py.sample
```diff
@@ -0,0 +1,10 @@
+COMMENT_SYSTEM = "MastodonComments"
```
Kwpolska, line 1: Consider using all-lowercase names.

v8/mastodoncomments/conf.py.sample
```diff
@@ -0,0 +1,10 @@
...
+COMMENT_SYSTEM_ID = "<YOUR-SITE_NAME>"
```
Kwpolska, line 2: Is this needed/used?

v8/mastodoncomments/README.md
MasinAD: Sure. I took localsearch as a template and got the order of `files` and `assets` wrong :-D. Maybe you want to take a look at that, too, then, for the sake of standardization.

v8/mastodoncomments/conf.py.sample
MasinAD: Will do.

v8/mastodoncomments/conf.py.sample
MasinAD: Actually, I didn't know what its original purpose was. I now read the Nikola documentation and found out that the COMMENT_SYSTEM_ID is used by several comment systems for specifying something like the next step in the process of accessing the comments.

In the light of this I will remove mastodon_config['host'] and use comment_system_id in its place. Until there's a bigger redesign in how to configure the comment systems, e.g. in a designated dict, I think its best to stick to lived practice.

MasinAD:
Ready for another round.

v8/mastodoncomments/templates/jinja/comments_helper_mastodoncomments.tmpl
```diff
@@ -0,0 +1,31 @@
...
+
+<p>With an account on the Fediverse or Mastodon, you can respond to this <a href="https://{{ comment_system_id }}/@{{ mastodon_config['account'] }}/{{ post.meta('mastodon') }}">post</a>. Since Mastodon is decentralized, you can use your existing account hosted by another Mastodon server or compatible platform if you don't have an account on this one. Known non-private replies are displayed below.</p>
+
+<p id="mastodon-comments-list"><button id="load-comment">Load comments</button></p>
+<div id="comments-wrapper">
+<noscript><p>Loading comments relies on JavaScript. Try enabling JavaScript and reloading, or visit <a href="https://{{ comment_system_id }}/@{{ mastodon_config['account'] }}/{{ post.meta('mastodon') }}">the original post</a> on Mastodon.</p></noscript>
+</div>
+<noscript>You need JavaScript to view the comments.</noscript>
+
+<script type="text/javascript" src="/js/purify.min.js"></script>
```
Kwpolska, line 13: Please adjust the templates and README to use the `/assets/js` and `/assets/css` folders.

Kwpolska: I want to use `<script type="text/javascript" src="/assets/js/purify.min.js"></script>` and `files/assets/js/purify.min.js` — so we keep all CSS and JS in the assets folder (as Nikola does).

MasinAD:
Changes done, please review. I overhauled the commands for installation a little bit, that's why there are several commits.

v8/mastodoncomments/mastodoncomments.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2021 Pieter David
```
Kwpolska, line 3: You probably want your name here :)

v8/mastodoncomments/files/assets/js/mastodon.sd.js
```diff
@@ -0,0 +1,144 @@
+const MASTODON_ACCOUNT_ID = '109270094542366763'
+const MASTODON_HOST = 'nerdculture.de'
```
Kwpolska, line 2: Should this be hardcoded here? Where are those `.sd.css` and `.sd.js` files used?

v8/mastodoncomments/mastodoncomments.py
MasinAD: Yeah, everybody should know who to blame. :-D

v8/mastodoncomments/files/assets/js/mastodon.sd.js
MasinAD: They aren't (anymore). They were the results of my first try using the method by sd.ai (mentioned somewhere as inspiration) but it wasn't indented on responses so I tried Carl Schwan's method. They are more similar than distinct as I learned during integration but hindsight is 20/20. I removed them. Maybe in the future I take another look because I like the automatic loading. I recognize that there are valid use-cases to not load the comments immediately so maybe that should be an option.

MasinAD:
We are getting closer to an end. Please review. :smile:

Kwpolska (approved):

Kwpolska:
Thanks for contributing!

MasinAD:
Thanks for all your advice. It was a pleasure.

PR #443, opened 2024-08-12
Rename the commonmark plugin module

CarstenGrohmann:
to prevent name collisions with the Python commonmark parser.

This change prevents:

```
AttributeError: module 'commonmark' has no attribute 'Parser'
```


Kwpolska (approved):

Kwpolska:
Thanks for contributing!

PR #452, opened 2025-01-16
Org-mode src block fixes

gudzpoz:
This PR contains two simple fixes for the org-mode plugin:

1. Use pygments' Emacs Lisp highlighting instead of Scheme highlighting for ELisp, since pygments has supported it for years (https://github.com/pygments/pygments/blame/edef94d66c2d70f05a86ac6098a69ab253b8d946/pygments/lexers/lisp.py#L1668).
2. Use `org-export-unravel-code` to extract code from org-mode source blocks to correctly handle ["any reference, protective commas and spurious indentation"](https://orgmode.org/worg/dev/org-export-reference.html#unravel-code). (The code now uses `define-advice` to reuse the original `org-html-src-block` as much as possible.)

(Also, there are some indentation adjustments when I re-format the code with vanilla Emacs.)


gudzpoz:
I added a few more changes to:

- Detect languages supported by Pygments ([`--json` option was added in Pygments four years ago](https://github.com/pygments/pygments/blame/edef94d66c2d70f05a86ac6098a69ab253b8d946/pygments/cmdline.py#L138)).
- Use [`pygmentize -g`](https://github.com/pygments/pygments/blame/edef94d66c2d70f05a86ac6098a69ab253b8d946/pygments/cmdline.py#L365) to detect the language when we really cannot find a matching language (and the user can specify `#+begin_src text` when they want un-highlighted texts, as is also expected in Emacs).

PR #454, opened 2025-02-12
Correct the name of wordpress_compiler plugin in metadata

evgeni:
This allows the plugin to be updated by `nikola plugin --upgrade`


felixfontein (approved):
Thanks for fixing this!

PR #456, opened 2025-02-17
Support file scheme and use bs4 to workaround missing 'body' element

jirib:
Support file scheme and use bs4 to workaround missing 'body' element


v8/import_page/import_page.py
```diff
@@ -61,34 +60,52 @@ class CommandImportPage(Command):
...
-            if sys.version_info[0] == 2 and isinstance(title, str):
-                title = title.decode('utf-8')
+        parse = requests.utils.urlparse(url)
+        if 'http' in parse.scheme:
+            r = requests.get(url)
+            if not (199 < r.status_code < 300):  # Did not get it
+                LOGGER.error(f'Error fetching URL: {url}')
+                return 1
+            html = r.content.decode(r.encoding).encode('utf-8') if r.encoding and 'utf-8' \
+                not in r.encoding.lower() else r.content
```
Kwpolska, line 76: This seems suspicious, why does `html` need to be encoded to `utf-8`?

jirib: Thank you; it was a remnant from my previous attempt to figure out why a page without 'body' (having 'ISO-8859-1' encoding) did produce empty list for 'nodes'.

v8/import_page/import_page.py
```diff
@@ -61,34 +60,51 @@ class CommandImportPage(Command):
...
+        if candidates:
+            node = max(candidates, key=lambda n: len(n.get_text(strip=True)))
         else:
-            LOGGER.error('Error fetching URL: {}'.format(url))
+            node = None  # empty
+
+        document = doc_template.format(
+            title=title,
+            slug=slug,
+            content=node.prettify()
```
jirib, line 107: Should we leave the HTML as it is? If it is 'article' should I just get 'article' content (mostly likely the website template would already have 'article' element) ???

jirib: I tried to add a functionality for this topic, see c7c9a55b7a340d9fc2235fcffe48becf257faa77 .

v8/import_page/import_page.py
```diff
@@ -0,0 +1,147 @@
...
+            utils.req_missing(['bs4'], 'use the import_page plugin')
+
+        while args:
+            arg = args.pop(0)
+            if arg == "-s" and args:
+                selector = args.pop(0)
+            elif arg == "-e" and args:
+                extractor = args.pop(0)
+            else:
+                urls.append(arg)  # Assume it's a page URL
```
Kwpolska, line 77: You don’t need to parse args yourself, you should use the built-in support in doit. See just about any command  plugin for an example.

v8/import_page/import_page.py
```diff
@@ -0,0 +1,147 @@
...
+
+from nikola.plugin_categories import Command
+from nikola import utils
+
+LOGGER = utils.get_logger('import_page', utils.STDERR_HANDLER)
+
+args = sys.argv[1:]
+selector = None # 'body'
+extractor = None # 'lambda node: BeautifulSoup(node.decode_contents(), "html.parser").prettify()'
+urls = []
```
Kwpolska, line 46: Is this used?

v8/import_page/import_page.py
```diff
@@ -0,0 +1,147 @@
...
+
+doc_template = '''<!--
+.. title: {title}
+.. slug: {slug}
```
Kwpolska, line 50: Consider also adding `date` (defaulting to now is fine).

v8/import_page/import_page.py
Kwpolska: I think we should remove the wrapper element (e.g. `<article>`) by default.

v8/import_page/import_page.py
jirib: I will come with a better solution, `extractor_module` so one can use an external file with its own code.

v8/import_page/import_page.py
jirib: IIUC all plugins just use `args` in `_execute()` as "inputs" for its processing. But i'd like to introduce an optional option. That is:

``` shell
<plugin> [-s extractor_file] arg...
```

That is why I was parsing it... Or, do I misunderstand your comment?

PR #459, opened 2025-08-29
wordpress_compiler: fix regex and plugin category

felixfontein:
The regex fix is necessary due to https://github.com/mrabarnett/mrab-regex/issues/584.

I've tried to understand what `(\s|\S|\Z)` should actually (other than matching a letter of anything or the end of the input, i.e. basically anything), and the conclusion seems to be that it's completely unnecessary. I also looked at the WordPress sources and found https://github.com/wordPress/wordPress/commit/d04cd4147aa3f74125e56f3d17b573393b331823, which basically does the same replacement as in this PR (namely, get rid of the bracket).


Kwpolska (approved):

felixfontein:
@Kwpolska thanks again! :)

PR #461, opened 2026-01-09
enable content search in flexsearch plugin

evgeni:
this was enabled in the past and then accidentally disabled

Fixes: 1f1975fa15bf15dd1840e236e55388971066f288

### DIFF
--- a/v8/flexsearch_plugin/flexsearch_plugin.py
+++ b/v8/flexsearch_plugin/flexsearch_plugin.py
@@ -61,7 +61,7 @@ class FlexSearchPlugin(LateTask):
                 if item.is_post and index_posts:
                     index[item.meta('slug')] = {
                         'title': item.title(),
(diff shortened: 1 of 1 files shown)


felixfontein (approved):
I assume you tested this?

evgeni:
Of course! ;)

(I did, really!)

evgeni:
the commit that disabled that says "- Commented out content indexing in plugin code but added comprehensive content indexing in JavaScript", but this doesn't seem to work -- as there is no "content" in the generated json, only when you enable the plugin code to add content, then json has it and JS can search

felixfontein:
In the same code in 1f1975fa15bf15dd1840e236e55388971066f288 there's also a lot of usage of `data[key].content`, so it doesn't look right that `content` isn't put into the JSON.

felixfontein:
Thanks for fixing this @evgeni!

